PR #68, opened 2021-06-28
Refactor code

ilyastrodubtsev:


robputt (approved):

robputt:
This looks sane enough, but I don't have a device to functionally test...

diagnostics-program/main.py
```diff
@@ -27,269 +26,131 @@
...
+
+
+def set_param_miner_diag(diagnostics):
+    param_miner_diag = utils.get_miner_diagnostics()
+    keys = ['MC', 'MD', 'MH', 'MN']
+    for (param, key) in zip(param_miner_diag, keys):
+        diagnostics[key] = param
+
+
+def write_info_to_files(prod_diagnostics, diagnostics):
```
vpetersson, line 94: While this is very compact and efficiently written, it's not super readable in my opinion. The *order* of the mapping between `path_list` and `data_list` is critical here. If the order in any of these were to change, it would break things. Maybe a dict (`{'data_json': '/opt/nebraDiagnostics/html/diagnostics.json'}`) would make this more readable. Moreover, there is no test if this order were to get shuffled.

vpetersson (approved):
LGTM

PR #72, opened 2021-07-04
Adds local development environment

vpetersson:
* Generic code cleanup
* Begins the de-camelCase process
* Adds more sensible defaults
* Adds support for running on a regular Raspberry Pi


robputt (approved):
We should run the test with the coverage option so we can easily see coverage percentage in the Github Actions logs, other than this looks good.

vpetersson:
> We should run the test with the coverage option so we can easily see coverage percentage in the Github Actions logs, other than this looks good.

Yeah good point, but one thing at the time.

Issue #61, opened 2021-06-16
Redesign diagnostics page


vpetersson:
Blocked by #60.

I want to give the diagnostics page a fresh new look.

Here's how it currently looks:
<img width="1679" alt="Screen Shot 2021-06-16 at 9 50 42 AM" src="https://user-images.githubusercontent.com/357664/122188910-590f9680-ce88-11eb-8c03-c679a7739cf9.png">

@vladimirpoleshko can you start mocking up a new design that goes in line with the dashboard design? Doesn't have to have the exact same design but at least have a similar feel to it to make it consistent.

shawaj:
Fixed in #92

Issue #60, opened 2021-06-16
Rewrite service to use Flask


vpetersson:
Let's rewrite the diagnostics page to use [Flask](https://flask.palletsprojects.com/en/2.0.x/), which gives us a lot of free features out of the box. We should then serve this app using [Gunicorn](https://gunicorn.org/).

To keep things simple, let's ditch the Nginx container for the time being and serve Gunicorn on port 8000 with Docker exposing it on port 80. That way we don't need to run the container as root.

This is a per-requisite for #39.

ryanteck:
@vpetersson Primarily the reason the container is run as root is to access many of the hardware and networking services. Not for Nginx.

robputt:
@vpetersson I'd quite like to claim this one, after looking at the code base + tests in the outstanding PR the conversion should be relatively straight forward.

shawaj:
Closed by #92

PR #92, opened 2021-08-04
Convert to Flask App

robputt:


hw_diag/utilities/hardware.py
```diff
@@ -0,0 +1,118 @@
...
+    return file.readline().strip().upper()
+
+
+def lora_module_test():
+    """
+    Checks the status of the lore module.
+    Returns true or false.
+    """
+    result = None
+    while result is None:
```
marvinmarnold, line 104: Couldn't this get stuck in an infinite loop if pktfwd never loads?

marvinmarnold: Can we use Python's native retry logic? https://julien.danjou.info/python-retrying/

hw_diag/utilities/miner.py
```diff
@@ -0,0 +1,77 @@
...
+    try:
+        miner_bus = dbus.SystemBus()
+        miner_object = miner_bus.get_object('com.helium.Miner', '/')
+        miner_interface = dbus.Interface(miner_object, 'com.helium.Miner')
+        try:
+            p2pstatus = miner_interface.P2PStatus()
+            param_list = [
+                str(p2pstatus[0][1]),
+                str(p2pstatus[1][1]),
+                str(p2pstatus[3][1]),
```
marvinmarnold, line 42: Just checking that this ordering is intentional. Is [this](https://github.com/helium/miner/blob/master/src/miner.erl#L175) the right place to look for the source data format?

hw_diag/utilities/shell.py
```diff
@@ -0,0 +1,41 @@
...
+        'BALENA_DEVICE_NAME_AT_INIT',
+        'BALENA_DEVICE_UUID',
+        'BALENA_APP_NAME',
+        'FREQ',
+        'FIRMWARE_VERSION',
+        'VARIANT'
+    ]
+    keys = ["BN", "ID", "BA", "FR", "FW", "VA"]
+
+    for (var, key) in zip(env_var, keys):
```
marvinmarnold, line 18: We should probably create a dictionary and iterate over that

env_var_mappings = {
  'BALENA_DEVICE_NAME_AT_INIT': 'BN',
  ...
}

But really, why not just use the env var name as the key on diagnostics? I made the same comment earlier. Is there a benefit in renaming the keys to be shorter?

hw_diag/utilities/variant_definitions.py
```diff
@@ -0,0 +1,215 @@
+variant_definitions = {
```
marvinmarnold, line 1: This is being copied from [another library](https://raw.githubusercontent.com/NebraLtd/helium-hardware-definitions/master/variant_definitions.py). Following Flask conventions, is this the right place? I've been putting this in a lib/` folder.

hw_diag/views/diagnostics.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@DIAGNOSTICS.route('/initFile.txt')
+def get_initialisation_file():
+    diagnostics = read_diagnostics_file()
+
+    if diagnostics.get('error'):
+        return 'Internal Server Error', 500
+
+    response = {
```
marvinmarnold, line 41: Why do we return a subset of all information?

marvinmarnold (dismissed):
I left a lot of comments but this is a big improvement over what's there. I know many of my comments address pre-existing issues. We don't need to address all of the things but I think we can improve readability a bit. At least a few tests should also be added.

hw_diag/views/diagnostics.py
robputt: A requirement was to retain backwards compatibility with the old app, this is used to generate /initFile.txt and is used in the factory for device onboarding.

hw_diag/utilities/variant_definitions.py
robputt: Flask is mostly unopinionated about this kind of stuff, really this should be packaged and included via requirements.txt so it is easily re-usable in other projects, but thats out of scope for this story to convert the existing logic to Flask. I would suggest we create a story to find a better way on how to distribute and use this file.

hw_diag/utilities/shell.py
robputt: This code is lifted from the rework that Ilya completed previously. Not sure if there is any benefit in using a hash table over using 2 lists and the zip function, I didn't really feel the need to change it.

We cannot alter the keys as we need to retain backwards compatibility, a requirement from Viktor, as the factory tooling relies on this naming convention. Renaming these keys should be completed though, as they are ugly right now, but this is probably out of scope for a Flask conversion and needs to be a story for the future so we can ensure the factory tooling is updated.

hw_diag/utilities/hardware.py
robputt: It needs to load eventually, else. the miner is essentially totally broken, so if we wait forever or if we retry it. is kind of irrelevant at that point.

hw_diag/utilities/hardware.py
```diff
@@ -0,0 +1,118 @@
...
+    keys = ["ECC"]
+
+    for (command, param, key) in zip(commands, parameters, keys):
+        try:
+            diagnostics[key] = config_search_param(command, param)
+        except Exception as e:
+            logging.error(e)
+
+
+def get_ethernet_addresses(diagnostics):
```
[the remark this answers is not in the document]
robputt: Yep this is a historic thing, remember https://github.com/NebraLtd/hm-diag/issues/60 is all about converting this to a Flask app, not a larger refactor of other elements of the code base. These files are largely moved from the previous implementation. We should split reworking this into a separate story.

hw_diag/utilities/hardware.py
```diff
@@ -0,0 +1,118 @@
+import logging
+import os
+from time import sleep
+
+from hw_diag.utilities.shell import config_search_param
+
+
+def detect_bt_lte(diagnostics):
```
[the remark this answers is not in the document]
robputt: Renamed...

hw_diag/utilities/blockchain.py
```diff
@@ -0,0 +1,23 @@
...
+        result = result.json()
+        try:
+            result = result['data']['height']
+        except KeyError:
+            raise KeyError(
+                "Not found value from key ['data']['height'] in json"
+            )
+        return result
+    else:
+        return "1"
```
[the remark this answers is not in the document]
robputt: Now returns None on error.

hw_diag/templates/diagnostics_page.html
```diff
@@ -0,0 +1,152 @@
...
+                <tr>
+                  <td>BT Detected</td>
+                  {% if diagnostics.BT %}
+                    <td class="text-success text-right">True</td>
+                  {% else %}
+                    <td class="text-danger text-right">False</td>
+                  {% endif %}
+                </tr>
+                <tr>
+                  <td>LoRa OK?</td>
```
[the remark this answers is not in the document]
robputt: See https://github.com/NebraLtd/hm-diag/issues/61

hw_diag/tasks.py
```diff
@@ -0,0 +1,110 @@
...
+    set_param_miner_diag(diagnostics)
+
+    # If  symmetric NAT set miner to relayed.
+    if diagnostics['MN'] == "symmetric":
+        diagnostics['MR'] = True
+    else:
+        diagnostics['MR'] = False
+
+    # Get the blockchain height from the Helium API
+    value = "1"
```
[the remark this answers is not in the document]
robputt: Unsure, this is historic code, if in doubt leave it in so it doesn't break stuff, prefer not to perform scream tests.

hw_diag/tasks.py
```diff
@@ -0,0 +1,110 @@
...
+    get_rpi_serial(diagnostics)
+    detect_ecc(diagnostics)
+
+    diagnostics['LOR'] = lora_module_test()
+
+    # Get the Public Key, Onboarding Key & Helium Animal Name
+    try:
+        data = get_public_keys()
+    except PermissionError as e:
+        data = None
```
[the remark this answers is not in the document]
robputt: With it outside of the try block None still gets written to the file. With it in the try block the file does not get written.

hw_diag/tasks.py
```diff
@@ -0,0 +1,110 @@
...
+
+    now = datetime.datetime.utcnow()
+    diagnostics['last_updated'] = now.strftime("%H:%M UTC %d %b %Y")
+
+    get_ethernet_addresses(diagnostics)
+    get_environment_var(diagnostics)
+    get_rpi_serial(diagnostics)
+    detect_ecc(diagnostics)
+
+    diagnostics['LOR'] = lora_module_test()
```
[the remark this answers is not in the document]
robputt: Nope, we must retain this structure as the factory tooling uses it :-(.

hw_diag/app.py
```diff
@@ -0,0 +1,51 @@
...
+def main():
+    app = get_app(__name__)
+    debug = False
+    if DEBUG:
+        debug = True
+    app.run('0.0.0.0', threaded=True, debug=debug)
+
+
+if __name__ == '__main__':
+    main()
```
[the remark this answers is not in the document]
robputt: Removed

hw_diag/tasks.py
```diff
@@ -0,0 +1,110 @@
...
+from hw_diag.utilities.miner import set_param_miner_diag
+from hw_diag.utilities.shell import get_environment_var
+from hw_diag.utilities.variant_definitions import variant_definitions
+
+
+log = logging.getLogger()
+log.setLevel(logging.DEBUG)
+
+
+def perform_hw_diagnostics():
```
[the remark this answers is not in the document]
robputt: In my opinion out of scope for this PR, this function is lifted from previous version and wrapped in APScheduler. The aim of this PR is to convert to Flask not massively refactor the other business logic. We should tackle this in a separate story.

hw_diag/utilities/hardware.py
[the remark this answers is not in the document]
marvinmarnold: Added a requirement to rename all other methods consistently in https://github.com/NebraLtd/hm-diag/issues/95#issue-961891227

hw_diag/utilities/hardware.py
marvinmarnold: In that case, wouldn't we still want to display the diagnostics page but show that there's a serious error? The way it is now, the program will block forever.

hw_diag/utilities/variant_definitions.py
marvinmarnold: Agreed: https://github.com/NebraLtd/helium-hardware-definitions/issues/1

marvinmarnold (commented):
Looking real good with all those changes. I'd still like to see some tests added as part of this work. If you don't think you've added any net new logic that needs to be tested, then could you add a handful of tests for pre-existing logic to establish a pattern for how Flask tests should be done?

Could you also add some documentation? Should include how to run tests. Also if there is a recommended way to run the site locally during development.

hw_diag/utilities/hardware.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def set_diagnostics_bt_lte(diagnostics):
+    devices = {
+        'BT': '0a12',
+        'LTE': '2c7c'
+    }
+
+    for dev_type, dev_addr in devices.items():
+        resp = os.popen('grep 0a12 /sys/bus/usb/devices/*/idVendor').read()
```
shawaj, line 15: shouldn't this have a variable instead of just `0a12` after the `grep`?

As currently it won't find the LTE will it?

hw_diag/utilities/hardware.py
robputt: Handle as bug seperately.

hw_diag/tasks.py
```diff
@@ -36,11 +36,10 @@ def perform_hw_diagnostics():
```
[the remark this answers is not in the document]
robputt: https://github.com/NebraLtd/Hotspot-Production-Tool/blob/main/main.py#L199

hw_diag/utilities/hardware.py
robputt: Thanks @shawaj nice catch I'll get this sorted.

hw_diag/utilities/variant_definitions.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # Nebra Outdoor Hotspot Gen1
+    'NEBHNT-OUT1': {
+        'FRIENDLY': 'Nebra Outdoor Hotspot Gen 1',
+        'APPNAME': 'Outdoor',
+        'SPIBUS': 'spidev1.2',
+        'RESET': 38,
+        'MAC': 'eth0',
+        'STATUS': 25,
+        'BUTTON': 26,
```
shawaj, line 36: BUTTON needs changing to 24

See https://github.com/NebraLtd/hm-pktfwd/pull/31

hw_diag/utilities/variant_definitions.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # Nebra Outdoor Hotspot Old Identifier
+    'Outdoor': {
+        'FRIENDLY': 'Nebra Outdoor Hotspot Gen 1',
+        'APPNAME': 'Outdoor',
+        'SPIBUS': 'spidev1.2',
+        'RESET': 38,
+        'MAC': 'eth0',
+        'STATUS': 25,
+        'BUTTON': 26,
```
shawaj, line 49: BUTTON needs changing to 24

See https://github.com/NebraLtd/hm-pktfwd/pull/31

shawaj (changes requested):

hw_diag/utilities/variant_definitions.py
shawaj: Now packaged Enjoy https://pypi.org/project/hm-hardware-defs/

You want to use v0.1.4

See also https://github.com/NebraLtd/hm-pktfwd/pull/32

hw_diag/utilities/variant_definitions.py
shawaj: Now packaged Enjoy https://pypi.org/project/hm-hardware-defs/

You want to use v0.1.4

See also https://github.com/NebraLtd/hm-pktfwd/pull/32

shawaj:
Looks good to me @robputt - tested it on a couple of test devices and works nicely.

Few things I noticed, but pretty minor and can probably be fixed later I guess:
- on the backend logs it was complaining about not having a timezone set and saying it was a development server and not to use in production
- some of the font downloads were 404ing according to logs (nunito-regular.woff and nunito-bold.woff)
- when the region has been set with a REGION_OVERRIDE env variable instead of directly via the miner it gives a "region plan" of UN123 instead of what's in the region override. Not sure if this makes sense or not. For production we don't use an override but in testing of packet forwarder and light hotspots we sometimes do (especially with non location asserted devices) see [here in packet forwarder](https://github.com/NebraLtd/hm-pktfwd/blob/c12a73944d23bba92e57cdd87043d56fe0b37819/files/configurePktFwd.py#L47). Perhaps if there is a region override set, we should mention that in the diagnostics like `EU868 (Set by override)` and could do similar when set by miner perhaps.
- in the original diag, instead of showing UN123 directly it showed ["awaiting location assertion"](https://github.com/NebraLtd/hm-diag/blob/99f83a011bf1c016adcd6a7cc7272573f5925612/diagnostics-program/html_generator.py#L126)
- LTE detected shows even for non-outdoor units. I'm wondering whether we should add a field to the [hardware variants](https://github.com/NebraLtd/helium-hardware-definitions) of "LTE Support" or something to that effect? So that we can automatically determine whether it should even bother to check for LTE?
- on a side note, perhaps cellular detected or modem detected is better than LTE?
- could add a favicon perhaps?

Going to mark as approved anyway as I guess all this can be resolved in a later update.

shawaj (approved):

shawaj:
Some screenshots of it running on a RAK based light hotspot I tested on. Looks smart!

Screenshot_20210814-195622_Chrome
Screenshot_20210814-195630_Chrome
Screenshot_20210814-195639_Chrome

robputt:
> Looks good to me @robputt - tested it on a couple of test devices and works nicely.
> 
> Few things I noticed, but pretty minor and can probably be fixed later I guess:
> 
> * on the backend logs it was complaining about not having a timezone set and saying it was a development server and not to use in production
> * some of the font downloads were 404ing according to logs (nunito-regular.woff and nunito-bold.woff)
> * seems to download the font every time the page re-loads automatically and therefore isn't instantly applied. Could we just cache it locally instead or bundle the font or something?
> * when the region has been set with a REGION_OVERRIDE env variable instead of directly via the miner it gives a "region plan" of UN123 instead of what's in the region override. Not sure if this makes sense or not. For production we don't use an override but in testing of packet forwarder and light hotspots we sometimes do (especially with non location asserted devices) see [here in packet forwarder](https://github.com/NebraLtd/hm-pktfwd/blob/c12a73944d23bba92e57cdd87043d56fe0b37819/files/configurePktFwd.py#L47). Perhaps if there is a region override set, we should mention that in the diagnostics like `EU868 (Set by override)` and could do similar when set by miner perhaps.
> * in the original diag, instead of showing UN123 directly it showed ["awaiting location assertion"](https://github.com/NebraLtd/hm-diag/blob/99f83a011bf1c016adcd6a7cc7272573f5925612/diagnostics-program/html_generator.py#L126)
> * LTE detected shows even for non-outdoor units. I'm wondering whether we should add a field to the [hardware variants](https://github.com/NebraLtd/helium-hardware-definitions) of "LTE Support" or something to that effect? So that we can automatically determine whether it should even bother to check for LTE?
> * on a side note, perhaps cellular detected or modem detected is better than LTE?
> * could add a favicon perhaps?
> 
> Going to mark as approved anyway as I guess all this can be resolved in a later update.


* QUESTION - The TZ issue is a Dockerism as Docker containers do not have a time zone set usually and hence default to UTC. I personally like UTC for machines as it makes them easily comparable if they are all in 1 universal time zone, however if you prefer to set a time zone please let me know which TZ and we'll create an issue for future enhancement for TZ support. The current deployed version also defaults to UTC it just doesn't scream about it.
* QUESTION - Using development server - currently the container just runs the CherryPi Python HTTP server as I didn't feel the need to bloat with NGINX / Apache, however as per the warning it is not a production ready server, mostly in terms of scaling although this should be a moot point as only a few requests are every going to hit these endpoints and they *should* be internal to the customers network. Although if you prefer to use WSGI and use a "real" web server I can edit the docker accordingly.
* NEW ISSUE -  Add caching headers for static files to prevent re-download each page refresh -  see https://github.com/NebraLtd/hm-diag/issues/102
* NEW ISSUE -  Favicon - see https://github.com/NebraLtd/hm-diag/issues/100
* FIXED - LTE detected now says "Modem Detected" in the template as requested
* NEW ISSUE - LTE showing on both indoor and outdoor variants - see https://github.com/NebraLtd/hm-diag/issues/101 (this is a regression from the previous version although I believe it should be handled separately as this has now been added to the variants file and will require some rework.
* FIXED - UN123 should show "Awaiting Location Assertion"

PR #93, opened 2021-08-04
docs: add beginner instructions to README

marvinmarnold:
**Why**
As a developer without any prior knowledge, I should be able to get started without reading any code.

**How**
Explain how to access the web interface and how to deploy to a development device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,22 @@ Because the stack is tightly intertwined with Balena, the easiest way to test th
     * Click `Download Balena OS`
 * Use [Etcher](https://www.balena.io/etcher/) to flash the downloaded image
 * Insert flash drive into the Raspberry Pi and boot (don't forget to plugin ethernet if necessary)
+* Set env vars for the application in Balena:
+    * `FREQ`: 868, 915, etc.
+    * `VARIANT`: Choose from [here](https://raw.githubusercontent.com/NebraLtd/helium-hardware-definitions/master/variant_definitions.py)
 * Deploy changes to:
     * All devices in application: `balena push BALENA_APPLICATION`
-    * Single device in local mode: `balena push UUID.local`
+    * Single device in local mode: `balena push UUID.local` (this will build on the device and )
 
 If you are on the same network as the Raspberry Pi, enter `LOCAL IP ADDRESS` from Balena into the browser.
 
 ### Deprecated deployment
+This is no longer [the recommended way](https://www.balena.io/docs/learn/deploy/deployment/#overview) of doing Balena deployments.
+
 * Add the remote Balena repo:`git remote add balena BALENA_USERNAME@git.balena-cloud.com:BALENA_USERNAME/BALENA_PROJECT.git`
 * Deploy changes: `git push balena YourLocalBranch:master`
+
+## Access from other networks
+
+Balena will generate a public URL for a device if [PUBLIC DEVICE URL](https://www.balena.io/docs/learn/manage/actions/#enable-public-device-url)
+toggled from the Balena device dashboard. This is not generally recommended, except for debugging.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,16 @@
 # hm-diag
 Helium Miner Diagnostics
 
-# Diagnostics JSON Layout
+**hm-diag** is a small website that displays diagnostic information about a hotspot.
+The website is only accessible if you are on the same network as the hotspot.
```
marvinmarnold, line 5: Is access really restricted? I haven't dug through to find how we accomplish this.

README.md
```diff
@@ -40,17 +49,23 @@ As part of the code the system produces a JSON file which then is used to carry
...
+* Add a device:
+    * Select newest version
+    * Production
+    * Click `Download Balena OS`
+* Use [Etcher](https://www.balena.io/etcher/) to flash the downloaded image.
+* Insert flash drive into the Raspberry Pi and boot (don't forget to plugin ethernet if necessary).
+* Deploy changes: `balena push BALENA_PROJECT_OR_DEVICE`
 
-You can now push your changes using the following command:
+If you are on the same network as the Raspberry Pi, enter `LOCAL IP ADDRESS` from Balena into the browser.
```
marvinmarnold, line 67: Does anybody have instructions for connecting remotely handy?

README.md
robputt: No access restriction, some people have forwarded port 80 to these and have it exposed on the internet -  not advised (https://www.shodan.io/search?query=title%3A%22Nebra+Hotspot+Diagnostics%22)

README.md
robputt: We probably don't want to connect remotely, it's designed to be a local debug / status tool, for remote operation see https://dashboard.nebra.com

robputt (approved):

README.md
marvinmarnold: Our support team never has to do this @vpetersson? It would be nice to be able to leave something plugged in at my coworking space and debug from home. I get that we don't want them to be publicly accessible by default but I'm guessing there's a way to connect remotely if you have Balena access.

README.md
marvinmarnold: Thanks, added a clarification: https://github.com/NebraLtd/hm-diag/pull/93/commits/a0fa55898a5c9c491c6dbb7d5b6624aa3f8e2cfa

README.md
shawaj: It's possible through balena yes. Using their PUBLIC DEVICE URL toggle

https://www.balena.io/docs/learn/manage/actions/#enable-public-device-url

shawaj: Also, one thing to note is you may need some environment variables to not throw errors. I'm not 100% sure but the to get the frequency I think it needs FREQ set to 868 or whatever it is. And the VARIANT set to whatever hardware it is from the hardware definitions repo

marvinmarnold: Thanks for pointing that out. [These](https://github.com/NebraLtd/hm-diag/blob/5436da7e8a06cae35791cf00381abe77ed4885f2/diagnostics-program/main.py#L52) are all the env vars that the app references. I'll double check which ones are automatically being set.

```
[
        'BALENA_DEVICE_NAME_AT_INIT',
        'BALENA_DEVICE_UUID',
        'BALENA_APP_NAME',
        'FREQ',
        'FIRMWARE_VERSION',
        'VARIANT'
    ]
```

robputt (approved):

README.md
shawaj: I think freq and variant come from the balena app environment variable settings.

The firmware version is from the docker compose.

Guess the others are automatically pulled in from balena

vpetersson: @marvinmarnold all the ones with the `BALENA_*` prefix should be set automatically from Balena. The rest I believe we set either on a per-device level (such as in TestNet) or in the App (the production deployments).

Issue #104, opened 2021-08-16
Sync percentage rounding


KevinWassermann94:
The sync percentage should only have 2 decimals and should round up to 100%, if higher than 99.99% as customers might think it is not fully synced.

KevinWassermann94:
And to expand this the sync percentage can go above 100 as seen in this picture:
![image](https://user-images.githubusercontent.com/82943687/129576786-67f30a65-37be-429a-b511-98f4816caee7.png)

shawaj:
This relates to #71 as well

I was trying to get clarity from helium on what actually constitutes synced in https://github.com/helium/miner/issues/957 but haven't had an answer yet.

In the explorer they call within 1500 blocks synced now but that's just an approximation. As the diagnostics is more real time I think we can be more accurate with this but not sure

Issue #71, opened 2021-07-03
Add "synced" call to show all ok


shawaj:
Currently we only show the sync percentage:
![Screenshot_20210703-091908_Chrome](https://user-images.githubusercontent.com/3359418/124348293-0bfa2700-dbe1-11eb-961d-56baf6cefc78.jpg)

This can be confusing to people new to Helium as it's actually considered synced if it's within 500 blocks and this is how Helium display it on the explorer as shown here:
https://github.com/helium/explorer/blob/6c5539bc121b49f443838fad811d35100c3772a9/components/Hotspots/StatusPill.js#L33

We should add something similar so that we don't have people worrying that they aren't quite synced

Maybe "99.999% (Synced within 500 blocks)" or similar

shawaj:
Seee https://github.com/helium/miner/issues/957 and https://github.com/NebraLtd/hm-diag/issues/104#issuecomment-899654158 as well

shawaj:
From the other thread....

> Essentialy, you're synced enough to earn if you're close enough to submit poc requests before they're too old, and will earn more if you're synced enough to get and process requests to you before they time out. You'll also need to be synced enough to know who the current consensus group is so that you can submit transactions. Typically this will mean you need to be within maybe 250 blocks of the newest block, but it's not impossible to be a little farther back and still be able to make things happen.

PR #107, opened 2021-08-16
Format sync percentage as per Issue #104

robputt:
Format sync percentage as described in #104 

- If greater than 99.8% consider the miner as synced and display 100%
- Format float as 2 decimal places
- Display 100% for percentages > 100%

Closes: #104

### DIFF
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -44,7 +44,11 @@
                 </tr>
                 <tr>
                   <td>Sync Percentage</td>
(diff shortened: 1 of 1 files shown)


shawaj:
IMO we should pull this in for now, but once we have more info on #71 re what is considered synced we should use that as the benchmark for what is displayed as 100%

shawaj (approved):

shawaj:
@robputt FYI https://github.com/NebraLtd/hm-diag/issues/71#issuecomment-899999836

Issue #103, opened 2021-08-16
Use some form of WSGI server


robputt:
Use some form of WSGI server for hosting the Flask app.

PR #109, opened 2021-08-16
Use gunicorn

robputt:
https://github.com/NebraLtd/hm-diag/issues/103

Convert container to serve via gunicorn.


vpetersson (approved):
LGTM

PR #113, opened 2021-08-17
Basic JSONRPC implementation

robputt:
Do not merge, this is just for testing / evaluation.


marvinmarnold:
This should ultimately be included in pyhelper so that there is a consistent way that we access gateway info across projects.

marvinmarnold:
This is now being done in hm-pyhelper. Closed by: https://github.com/NebraLtd/hm-pyhelper/pull/8

Issue #41, opened 2021-05-31
Add more features / match with dashboard


shawaj:
I think it would be nice if the remote dashboard and the local dashboard converge to have similar features from a device info / device control perspective...

Things we could add:
- reboot, shutdown, restart
- relayed status
- network statistics 
- Eth / WiFi status
- WiFi network search and update
- mDNS / NetBIOS search (nebra.local or macaddress.local or similar)
- local / wan IP displayed for all interfaces
- explore miner features
  - peer book
  - connect to new peers
  - take snapshot
  - restore snapshot
- Customisable lora RF settings (rssi offset / antenna gain / tx power)

Probably a bunch more we could add, just wanted to get the conversation started 👌

ryanteck:
I mean kinda the idea of the diagnostics system is that it is for diagnostics and picking up on issues.

Adding all of the above then turns it more from a diagnostics system to a management panel. So is most likely an entire re-write of the system.

whittinghamj:
I really like these ideas.

Habing a local dashboard on the hotspot itself would be a massive help - maybe even have a little JSON API endpoing for the techie's to tap into as well.

I would ask that the API endpoint has a reboot option as well?

vpetersson:
Blocked by #39. However, once done, we should be able to knock most of these out fairly easily.

shawaj:
Some inspiration

https://medium.com/@syncrob_it/local-dashboard-customize-everything-5a57b129b112

shawaj:
![image](https://user-images.githubusercontent.com/3359418/122654410-1af1cb80-d143-11eb-8321-4091c9e5a70f.png)

shawaj:
Turns out balenaOS already has mDNS / NetBIOS enabled on `<shortuuid>.local` but we can modify this using the supervisor api

https://www.balena.io/docs/reference/supervisor/supervisor-api/#patch-v1devicehost-config

I'm thinking `nebra-<last 6 digits of Mac address>. local`

bjessie:
Could temperature be added to the local diagnostics page without much fuss? The metric is already available in the remote dashboard. I've had great difficulty keeping them cool and this would be incredibly helpful.

smurf12345:
[...]

PR #115, opened 2021-08-17
mdns: change hostname to something useful

shawaj:
changes hostname to nebra-<last 6 chars of mac>.local instead of balena uuid

Relates-to: #41
Signed-off-by: Aaron Shaw <shawaj@gmail.com>

**Why**

Make the diagnostics page easier to find locally, without knowing the local IP address of the unit.

**How**

Add a new function to set the hostname using the balena supervisor API

**References**

Part of #41

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,7 @@ services:
     labels:
       io.balena.features.sysfs: '1'
       io.balena.features.dbus: '1'
+      io.balena.features.supervisor-api: '1'
 
 volumes:
   miner-storage:
--- a/hw_diag/app.py
+++ b/hw_diag/app.py
@@ -5,6 +5,7 @@ from flask import Flask
 from flask_apscheduler import APScheduler
 
 from hw_diag.tasks import perform_hw_diagnostics
+from hw_diag.utilities.hostname import set_hostname
 from hw_diag.views.diagnostics import DIAGNOSTICS
 
 
@@ -32,6 +33,10 @@ def get_app(name):
     @scheduler.task('cron', id='run_diagnostics', minute='*/1')
     def run_diagnostics_task():
         perform_hw_diagnostics()
+
+    # Set hostname for mdns / netbios
+    def run_hostname_task():
+        set_hostname()
 
     # Register Blueprints
     app.register_blueprint(DIAGNOSTICS)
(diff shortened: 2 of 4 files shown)


hw_diag/app.py
```diff
@@ -33,6 +34,10 @@ def get_app(name):
     def run_diagnostics_task():
         perform_hw_diagnostics()
 
+    # Set hostname for mdns / netbios
+    def run_hostname_task():
```
shawaj, line 38: @robputt this works, but is probably in the wrong place? 

i guess we only need to run this once on first load of the container, rather than on a schedule.

Maybe there is a better way to do it?

hw_diag/utilities/hostname.py
```diff
@@ -0,0 +1,26 @@
+import os
+import requests
+
+from hw_diag.utilities.hardware import get_mac_addr
+
+def set_hostname():
```
shawaj, line 6: @robputt @vpetersson would we need to add a unit test for this function?

shawaj:
FYI this allows you to access the diagnostics as follows:
- mdns: http://nebra-xxxxxx.local
- netbios: http://nebra-xxxxxx
- ip: http://192.168.x.x (or whatever is it's local ip)

Where `xxxxxx` is the last 6 digits of the eth0 mac address

docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
     labels:
       io.balena.features.sysfs: '1'
       io.balena.features.dbus: '1'
+      io.balena.features.supervisor-api: '1'
```
vpetersson, line 22: Not sure how I feel about this. Exposing the supervisor API allows you to basically fully take over the device. We might need to tighten up the security first or what do you think @robputt?

shawaj: @vpetersson we could do it through the gateway-config container instead which already has the supervisor api enabled and isn't exposed externally?

Don't think there's any other way to do it without using the API?

vpetersson: Yes, doing this through the `gateway-config` container would be preferable from a security perspective as it's not publicly exposed. That said, I don't know why `gateway-config` would require access to Balena's Supervisor either. That's another component that we are yet to audit/rewrite.

shawaj: Either way, I think this is pretty essential. So one way or another we have to use the supervisor API to enable this.

shawaj:
Moved the non-mdns related changes to #122 so closing this now in favour of https://github.com/NebraLtd/hm-config/issues/62

PR #117, opened 2021-08-20
Fixes new layout of public_key file

vpetersson:
Hotfix for new layout of `public_keys` file

### DIFF
--- a/hw_diag/tests/test_get_public_keys.py
+++ b/hw_diag/tests/test_get_public_keys.py
@@ -6,14 +6,9 @@ from hw_diag.utilities.miner import get_public_keys  # noqa
 
 
 class TestGetPublicKeys(unittest.TestCase):
-    TEST_DATA = [
-        "{pubkey,",
-        '"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.',
-        '{onboarding_key,',
-        '"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.',
-        '{animal_name,',
-        '"gigantic-basil-turtle"}.'
-    ]
+    TEST_DATA = '{pubkey, "112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.'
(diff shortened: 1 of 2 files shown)


hw_diag/tests/test_get_public_keys.py
```diff
@@ -6,24 +6,21 @@
...
 class TestGetPublicKeys(unittest.TestCase):
-    TEST_DATA = [
-        "{pubkey,",
-        '"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.',
-        '{onboarding_key,',
-        '"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.',
-        '{animal_name,',
-        '"gigantic-basil-turtle"}.'
-    ]
+    TEST_DATA = '{}\n{}\n{}'.format(
```
marvinmarnold, line 9: I'm unsure about the line breaks. The miner I'm looking at doesn't appear to have any.

marvinmarnold (approved):
The code looks internally consistent. But I don't quite understand the context. Do you have a cross-link to the code that generates the `public_keys` file?

hw_diag/tests/test_get_public_keys.py
vpetersson: It was introduced in the latest GA. We've had half the fleet fall over because of this.

robputt (approved):

hw_diag/tests/test_get_public_keys.py
vpetersson:
```
/tmp/build # wc -l /var/data/public_keys
3 /var/data/public_keys
```

vpetersson:
> The code looks internally consistent. But I don't quite understand the context. Do you have a cross-link to the code that generates the `public_keys` file?

Yeah, so this is just really a band aid while we wait for @robputt to adopt the internal API. I really hate how we shell out and read/parse random files on disk. It's just asking for trouble.

shawaj:
> The code looks internally consistent. But I don't quite understand the context. Do you have a cross-link to the code that generates the `public_keys` file?

https://github.com/NebraLtd/hm-miner/blob/master/start-miner.sh and https://github.com/helium/miner/blob/master/src/miner_keys.erl

PR #124, opened 2021-08-23
Deprecate hm-upnp rep

### LINKED ISSUE #123: Move uPNP into hm-diag
Looking at [hm-upnp](https://github.com/NebraLtd/hm-upnp), it really doesn't do anything. We could spare a container (and the subsequent overhead) by just moving this into this repo.

vpetersson:
**Why**

Because we want to reduce the footprint. Closes #123.

**How**

Ports https://github.com/NebraLtd/hm-upnp almost line-by-line but with the help of new cron task executer

This should probably be rewritten later with something like [uPnPy](https://pypi.org/project/UPnPy/).

Will require changes to the main docker-compose file.

**References**

* https://github.com/NebraLtd/hm-diag/issues/123
* https://github.com/NebraLtd/hm-upnp

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ RUN apk add --no-cache \
     python3=3.8.10-r0 \
     i2c-tools=4.1-r3 \
     usbutils=012-r1 \
+    miniupnpc=2.1.20191224-r0 \
     py3-pip=20.1.1-r0 \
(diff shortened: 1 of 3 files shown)


vpetersson:
Closing out this one as `hm-diag` is not running in `network: host` mode, which is appears to be a requirement for `upnpc` to actually work.

Issue #135, opened 2021-08-28
Speed up the generation of initFile.txt


vpetersson:
Right now, the biggest bottleneck in manufacturing is that we sit around and wait for `/initFile.txt` for a significant amount of time, which slows down throughput significantly.

Here are some ideas on how we could potentially accomplish this:

* Introduce a 'first boot' mode, where we for instance create a file on disk (e.g. `.initialized`). If this file doesn't exist, we will basically delay the start of all containers other than hm-diag with the objective to serve `/initFile.txt` as quickly as possible
* Reshuffle the boot order such that we always delay the start of everything other than hm-diag
* Change how we get the keys and get them from the chip direct rather than from via the miner

Issue #137, opened 2021-09-02
Deprecate hm-gwmfr and move to "gateway-mfr-rs"


vpetersson:
Helium now offers a Rust re-write of the gateway configuration tool available [here](https://github.com/helium/gateway-mfr-rs). Not only is it much smaller and faster, it also generates JSON instead of Erlang-like output.

I've manually compiled it and tested it on one of the miners, and it looks very promising:

Getting the key:
```
./gateway_mfr key 0
{
  "key": "112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc",
  "name": "gigantic-basil-turtle",
  "slot": 0
}
```

Running the tests:
```
./gateway_mfr test
{
  "result": "pass",
  "tests": [
    {
      "output": "ok",
      "result": "pass",
      "test": "serial"
    },
    {
      "output": "ok",
      "result": "pass",
      "test": "zone_locked(data)"
    },
    {
      "output": "ok",
      "result": "pass",
      "test": "zone_locked(config)"
    },
    {
      "output": "ok",
      "result": "pass",
      "test": "slot_config(0..=15, ecc)"
    },
    {
      "output": "ok",
      "result": "pass",
      "test": "key_config(0..=15, ecc)"
    },
    {
      "output": "ok",
      "result": "pass",
      "test": "miner_key(0)"
    }
  ]
}
```

# Timings

## Rust version

```
/tmp # time ./gateway_mfr key 0
{
  "key": "112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc",
  "name": "gigantic-basil-turtle",
  "slot": 0
}
real    0m 0.07s
user    0m 0.00s
sys     0m 0.00s
```

## Erlang version

TBC

## From miner (what we do currently)

Note that this is is with Miner running, so Erlang's VM already in memory.

```
/opt/miner # time /opt/miner/bin/miner print_keys
{pubkey,"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.
{onboarding_key,"112mPWkGW55kcbQTgbtJvgAKMSTrEhHgavrdF1Cbu8FU85tTL4Nc"}.
{animal_name,"gigantic-basil-turtle"}.
real    0m 6.11s
user    0m 2.30s
sys     0m 1.32s
```

PR #138, opened 2021-09-03
Manufacturing tweaks

vpetersson:
* Speeds up manufacturing
* Removes duplicate installed python dependencies
* Exposes WiFi MAC to `initFile.txt`

References:
* Closes https://github.com/NebraLtd/hm-diag/issues/135
* Refs https://github.com/NebraLtd/hm-diag/issues/137

### DIFF
--- a/.github/workflows/compile-docker.yml
+++ b/.github/workflows/compile-docker.yml
@@ -3,43 +3,6 @@ name: Compile Docker Images
 on: [push]
 
 jobs:
-  rust-compile:
-    name: Build gateway-mfr-rs
-    runs-on: ubuntu-latest
-    steps:
-      - name: Install stable toolchain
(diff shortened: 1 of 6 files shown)


rxt-rob (approved):
Maybe an improvement would be to check the MAC addresses for the network interfaces are valid, could pick up on bad wifi usb modules or whatever during the manufacturing onboarding, however there is no regression on the current implementation so we should not block merging as a result.

robputt (approved):

robputt:
Maybe an improvement would be to check the MAC addresses for the network interfaces are valid, could pick up on bad wifi usb modules or whatever during the manufacturing onboarding, however there is no regression on the current implementation so we should not block merging as a result.

PR #149, opened 2021-09-15
change to session bus

shawaj:
**Why**

**How**

**References**

### DIFF
--- a/hw_diag/tests/test_get_miner_diag.py
+++ b/hw_diag/tests/test_get_miner_diag.py
@@ -8,10 +8,10 @@ from hw_diag.utilities.miner import get_miner_diagnostics # noqa
 
 
 class TestGetMinerDiag(unittest.TestCase):
-    @patch("dbus.SystemBus")
+    @patch("dbus.SessionBus")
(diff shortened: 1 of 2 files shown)


vpetersson:
@shawaj Can you merge in master?

vpetersson:
Also, i take it this is contingent on `docker-compose.yml` changes in helium-miner-changes?

shawaj:
@vpetersson yes exactly re docker-compose.yml

however, I have just been testing and we need to put in a wait_for_dbus() call like:

```
#!/bin/bash

function wait_for_dbus() {
	while true; do
		dbus-send --system \
			  --print-reply \
			  --dest=org.freedesktop.DBus \
			  /org/freedesktop/DBus \
			  org.freedesktop.DBus.ListNames

		if [ $? ]; then
			break;
		else
			sleep 0.1
		fi
	done

	echo "DBus is now accepting connections"
}
```

shawaj:
ok added in the necessary wait scripts now @vpetersson so ready to go whenever

robputt (approved):

PR #150, opened 2021-09-16
Adopt hm-pyhelper

vpetersson:
* First step towards adopting hm-pyhelper


requirements.txt
```diff
@@ -4,5 +4,6 @@ certifi==2021.5.30
 dbus-python==1.2.16
 gunicorn==20.1.0
 hm-hardware-defs==0.1.5
-requests==2.25.1
+hm-pyhelper==0.2
+requests==2.26.0
```
marvinmarnold, line 8: Out of curiosity, this upgrade wasn't required was it?

marvinmarnold (approved):

robputt (approved):

requirements.txt
vpetersson: It was because i set this as the minimum version in the helper.

PR #164, opened 2021-09-29
feat: do not exec diagnostics if gwmfr running

marvinmarnold:
**Why**
Some ECC chips are being corrupted during manufacturing. There may be a race condition between gwmfr and diag. We want to wait for gwmfr to complete before running diagnostics.

**How**
Write to an empty file (AKA touch) when gwmfr completes successfully. Wait for this file to be present before running diagnostics.

**References**
- Slack convo: https://pisupply.slack.com/archives/C024BNQ1Y6T/p1632912403459500
- gwmfr: https://github.com/NebraLtd/hm-gwmfr/pull/18
- helium-miner-software: https://github.com/NebraLtd/helium-miner-software/pull/155

### DIFF
--- a/hw_diag/views/diagnostics.py
+++ b/hw_diag/views/diagnostics.py
@@ -1,6 +1,6 @@
 import json
 import base64
-import os.path
+import os
 import logging
 
 from flask import Blueprint
--- a/hw_diag/views/diagnostics.py
+++ b/hw_diag/views/diagnostics.py
@@ -53,8 +53,10 @@ def get_diagnostics():
         display_lte=display_lte
     )
 
+
 def is_gwmfr_running():
     return not os.path.isfile(ECC_SUCCESSFUL_TOUCH_FILEPATH)
+
 
 @DIAGNOSTICS.route('/initFile.txt')
 def get_initialisation_file():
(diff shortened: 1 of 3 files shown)


vpetersson:
@marvinmarnold the things we need to hold back is the `/initFile.txt` generation. That is what Hotspot Production Tool is [hammering](https://github.com/NebraLtd/Hotspot-Production-Tool/blob/main/app/main.py#L422).

marvinmarnold:
@vpetersson does the check need to happen in both places? I think we want to stop Hotspot-Production-Tool from continuing and make sure diagnostics doesn't access ECC.

hw_diag/views/diagnostics.py
```diff
@@ -53,18 +53,21 @@ def get_diagnostics():
...
     """
     This needs to be generated as quickly as possible,
     so we bypass the regular timer.
     """
-    
+
     if is_gwmfr_running():
-        logging.info("hm-gwmfr is still runnning. initFile.txt will not be returned")
+        logging.info("hm-gwmfr is still runnning. \
+            initFile.txt will not be returned.")
```
marvinmarnold, line 70: Just tested this. It will add some undesirable spacing but we can fix later.

I'm not actually sure the correct syntax. Is it?

```
logging.info("A" +
   "B")?

vpetersson (approved):

hw_diag/views/diagnostics.py
vpetersson: i think that's fine. Worse case scenario the logging gets a newline.

PR #171, opened 2021-10-19
Adds docker healthcheck

vpetersson:
* Adds Docker Healtcheck to Dockerfile
* Removes deprecated function

## References
* https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck
* https://docs.docker.com/engine/reference/builder/#healthcheck

### DIFF
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -1,7 +1,6 @@
 import logging
 import datetime
 import json
-import os.path
 
 from hm_pyhelper.hardware_definitions import variant_definitions
 from hm_pyhelper.miner_param import get_ethernet_addresses


hw_diag/views/diagnostics.py
```diff
@@ -54,21 +54,13 @@ def get_diagnostics():
     )
 
 
-def is_gwmfr_running():
-    return not os.path.isfile(ECC_SUCCESSFUL_TOUCH_FILEPATH)
```
marvinmarnold, line 58: `ECC_SUCCESSFUL_TOUCH_FILEPATH` declaration can also be removed https://github.com/NebraLtd/hm-diag/pull/171/files#diff-04b1aca2fd0e3d6dca6921bc407e7e1cf9b6c93e823d1d31591fcfd6b398304aL22

marvinmarnold (approved):

PR #173, opened 2021-10-20
Run diagnostics more frequently

shawaj:
**Why**

Current functionality runs every 5 minutes between the 5th and 55th minutes of the hour. This means it can take up to 5 to run at first boot and at the turn of the hour will not run from x.55 until x+1.05 (10 minute gap)

Seems better to run this every minute instead

**How**

Change schedule

**References**

https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html

### DIFF
--- a/hw_diag/app.py
+++ b/hw_diag/app.py
@@ -40,7 +40,7 @@ def get_app(name):
     @scheduler.task(
         'cron',
         id='run_diagnostics',
-        minute='1,2,3,4,5,10,15,20,25,30,35,40,45,50,55')
+        minute='*')
     def run_diagnostics_task():
         perform_hw_diagnostics(ship=False)
 


hw_diag/app.py
```diff
@@ -40,7 +40,7 @@ def get_app(name):
     @scheduler.task(
         'cron',
         id='run_diagnostics',
-        minute='5,10,15,20,25,30,35,40,45,50,55')
+        minute='*')
```
vpetersson, line 43: No, this wont' work. The reason why we have these explicit minutes (every 5, except `:00`) is because we have a custom job at the top of the hour. Hence if we just add `*` this will change the behavior. 

I will chat with @robputt how to change this and avoid the scheduler all together for everything but the actual GCS job as we really don't need the scheduler now that we've reworked things a fair bit.

vpetersson:
Superseded by https://github.com/NebraLtd/hm-diag/issues/174

hw_diag/app.py
shawaj: @vpetersson the scheduler can run multiple jobs at the same time.

vpetersson: Sure but there is really no need for a scheduler anymore. We are wasting a lot of CPU cycles on this. Might as well remove it as per https://github.com/NebraLtd/hm-diag/issues/174. This will both speed it up too, as we don't need to wait for potentially 59 seconds for it to run.

Issue #174, opened 2021-10-20
Remove scheduler


vpetersson:
With the new rework, combined with the fact that we are also checking the status of diagnostics before starting other services, we should probably rework the way we run the diagnostics to make it more snappy.

Here's what I suggest:

* We remove the scheduler all together for serving diagnostics (and instead add a cache on success perhaps)
* We leave the GSC upload in the scheduler

vpetersson:
Initially, I thought it would be a good idea to remove the file cache entirely. However, after checking how long it takes to run this function, I changed my mind:

```
/srv # python -c 'from hw_diag.tasks import perform_hw_diagnostics
> 
> ^C
> python -c 'from hw_diag.tasks import perform_hw_diagnostics^C
> '
/bin/sh: python: not found
/srv # time python3 -c 'from hw_diag.tasks import perform_hw_diagnostics; perform_h
w_diagnostics()'
[...]
INFO:root:Diagnostics complete
real    0m 7.83s
user    0m 5.25s
sys     0m 0.83s
```

shawaj:
We probably can simplify `perform_hw_diagnostics` ... For example the animal name, helium address, hardware serial, frequency, variant are never going to change so only need to be checked once at first run.

Region plan, firmware version are not going to change without the container restarting.

Sync percentage, height status, miner connected to blockchain, relayed are probably the ones that need updating most regularly

ECC detected, lora operational, LTE detected, Bluetooth detected, WiFi Mac address, Ethernet Mac address are unlikely to change very often so could be on a less frequent time period

PR #175, opened 2021-10-20
Removes schedule for serving diagnostics_json

vpetersson:
* Removes scheduler for generating the diagnostics JSON to speed up the first load time
* Introduces caching for JSON as it is a rather slow process
* (Fixes DEBUG string parsing)

Closes https://github.com/NebraLtd/hm-diag/issues/174

### DIFF
--- a/hw_diag/tests/test_views_diagnostics.py
+++ b/hw_diag/tests/test_views_diagnostics.py
@@ -1,6 +1,7 @@
 import unittest
 import flask
 from werkzeug.datastructures import ImmutableMultiDict
+from unittest.mock import patch, mock_open
 
 
 # Test cases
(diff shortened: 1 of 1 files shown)


shawaj:
Will this still auto-reload the page each time it pulls in new info? 

Also, is there a reason to only do this every 5 minutes?

hw_diag/views/diagnostics.py
```diff
@@ -24,7 +28,16 @@
 
 def read_diagnostics_file():
     diagnostics = {}
+
     try:
+        # Re-generate the cache if it is older than cache_cutoff
```
robputt, line 33: Why do we use this rather than using a library such as Flask-Cache which can cache in memory and is only 1 line of code (decorating the function).

robputt (approved):
Approving as there is no regression to the functionality with this PR. See comment about caching in line for consideration. Potentially we could offer the user the opportunity to invalidate the cache if they wish to have fresher results? A future enhancement may be using JS to load the results asynchronously to prevent perceived long page load when cache is expired.

hw_diag/views/diagnostics.py
vpetersson: Good idea. Yeah, that's cleaner.

ilyastrodubtsev (approved):

Issue #167, opened 2021-10-12
'ECC tests failed' from Hotspot Production Tool


vpetersson:
We're getting a lot of `Got HTTP Status Code 500 with the message "b'ECC tests failed'` from Hotspot Production Tool still, even upgrading to the latest disk image from 2021-09-30. The above disk image includes the new lock, which either doesn't doesn't work or the error is elsewhere.

PR #191, opened 2021-10-25
Removes ECC test

vpetersson:
Refs https://github.com/NebraLtd/hm-diag/issues/167.

@shawaj is insisting we remove the ECC test. I'm not personally convinced this is a good idea.


shawaj:
Guessing we can close this now @marvinmarnold @vpetersson ?

vpetersson:
Yeah this is not necessary now with the refactoring.

Issue #194, opened 2021-10-26
diagnostics['OK'] = public_keys['key'] Type Error bool


KevinWassermann94:
26.10.21 15:16:34 (+0200)  diagnostics  DEBUG:flask_caching.backends.simplecache:get key 'view//' -> miss 
26.10.21 15:16:34 (+0200)  diagnostics  INFO:root:Running periodic hardware diagnostics
26.10.21 15:16:35 (+0200)  diagnostics  ERROR:root:gateway_mfr exited with a non-zero status
26.10.21 15:16:36 (+0200)  diagnostics  ERROR:hw_diag:Exception on / [GET]
26.10.21 15:16:36 (+0200)  diagnostics  Traceback (most recent call last):
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 2070, in wsgi_app
26.10.21 15:16:36 (+0200)  diagnostics      response = self.full_dispatch_request()
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1515, in full_dispatch_request
26.10.21 15:16:36 (+0200)  diagnostics      rv = self.handle_user_exception(e)
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1513, in full_dispatch_request
26.10.21 15:16:36 (+0200)  diagnostics      rv = self.dispatch_request()
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1499, in dispatch_request
26.10.21 15:16:36 (+0200)  diagnostics      return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/flask_caching/__init__.py", line 475, in decorated_function
26.10.21 15:16:36 (+0200)  diagnostics      rv = f(*args, **kwargs)
26.10.21 15:16:36 (+0200)  diagnostics    File "/usr/local/lib/python3.9/site-packages/hw_diag-1.0-py3.9.egg/hw_diag/views/diagnostics.py", line 54, in get_diagnostics
26.10.21 15:16:36 (+0200)  diagnostics      diagnostics = read_diagnostics_file()
[...]

Issue #181, opened 2021-10-21
diagnostics JSON initFile should be fail-safe


shawaj:
currently, if diagnostics can't get some info, the JSON just shows "diagnostics has not yet run"

this slows down loading of hm-config significantly.

we should allow the JSON to output with default values in missing fields, such as "Still loading, please try again later!" or something like that instead of just showing "diagnostics has not yet run" as the entire json output

as discussed with @marvinmarnold on slack

PR #201, opened 2021-10-29
refactor initFile and prevent 500s

marvinmarnold:
**Why**
The initFile should always return something, even if the miner is not fully operational. 

**How**
Created DiagnosticsReport and Diagnostic concepts for each test we do on the miner.
Each Diagnostic must add a value to the report and cannot cause the report to fail.

The [tests](https://github.com/NebraLtd/hm-diag/pull/201/files#diff-b4c69b275920e57eaa26df2f0192bfc5ab6201d3a1e3583ee590ff3509540963R46) are helpful for understanding what the final output looks like.

**Next steps**
- Add more helpful error messages. Eg still swallowing some info when keys fail
- Update Hotspot-Production-Tool according to this logic
- Move DiagnosticsReport and Diagnostic to hm-pyhelper.
- Remove/update methods in hm-pyhelper that accept `diagnostics` as a parameter.
- Update `perform_hw_diagnostics` to use the same pattern.

**References**
- Closes: https://github.com/NebraLtd/hm-diag/issues/181
- Related to: https://github.com/NebraLtd/Hotspot-Production-Tool/issues/9
- Fixes: https://github.com/NebraLtd/hm-diag/issues/194 and supercedes: https://github.com/NebraLtd/hm-diag/pull/199

### DIFF
--- a/hw_diag/diagnostics/bt_lte_diagnostic.py
+++ b/hw_diag/diagnostics/bt_lte_diagnostic.py
@@ -1,0 +1,71 @@
+import os
+from hm_pyhelper.diagnostics.diagnostic import Diagnostic
+
+
+DEVICE_TYPE_ADDRESS_MAPPINGS = [
+    {
+        'dev_type': 'BT',
+        'dev_addr': '0a12'
+    },
+    {
+        'dev_type': 'LTE',  # Quectel
+        'dev_addr': '2c7c'
+    },
+    {
+        'dev_type': 'LTE',  # Sierra Wireless MC7700
+        'dev_addr': '68a2'
+    },
+    {
+        'dev_type': 'LTE',  # Telit / Reyax
+        'dev_addr': '1bc7'
+    },
+    {
+        'dev_type': 'LTE',  # SimCom SIM7100E
+        'dev_addr': '1e0e'
+    },
+    {
+        'dev_type': 'LTE',  # Huawei ME909s-120
+        'dev_addr': '12d1'
+    },
+    {
+        'dev_type': 'LTE',  # MikroTik R11e-LTE6
+        'dev_addr': '2cd2'
+    }
+]
+
+
+class BtLteDiagnostic(Diagnostic):
+    def __init__(self, dev_type, dev_addr):
+        super(BtLteDiagnostic, self).__init__(dev_type, dev_type)
+        self.dev_addr = dev_addr
+
+    def perform_test(self, diagnostics_report):
+        resp = os.popen(
+            'grep %s /sys/bus/usb/devices/*/idVendor' % self.dev_addr
+        ).read()
+
+        if self.dev_addr in resp:
+            print("setting %s to %s" % (self.key, True))
+            diagnostics_report.record_result(True, self)
+        else:
+            # Only set the dev_type to False if it is not already true
+            if self.key not in diagnostics_report:
+                diagnostics_report.record_result(False, self)
+
+            dev_type_already_true = diagnostics_report[self.key]
+            if not dev_type_already_true:
+                diagnostics_report.record_result(False, self)
+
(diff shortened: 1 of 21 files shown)


hw_diag/views/diagnostics.py
```diff
@@ -68,52 +70,21 @@ def get_initialisation_file():
...
+        LoraDiagnostic(),
+        KeyDiagnostics(),
+        # Must be last, it depends on previous results
+        PfDiagnostic()
+    ]
+    diagnostics_report = DiagnosticsReport(diagnostics)
+    diagnostics_report.perform_diagnostics()
+
+    keys_to_extract = ["VA", "FR", "E0", "W0", "RPI", "OK", "PK", "PF", "ID"]
+    response = diagnostics_report.get_report_subset(keys_to_extract)
```
marvinmarnold, line 88: Is there harm in returning the full report here? I'd like to include at least the relevant friendly keys too. Ideally, the full report.

hw_diag/diagnostics/bt_lte_diagnostic.py
```diff
@@ -0,0 +1,71 @@
...
+]
+
+
+class BtLteDiagnostic(Diagnostic):
+    def __init__(self, dev_type, dev_addr):
+        super(BtLteDiagnostic, self).__init__(dev_type, dev_type)
+        self.dev_addr = dev_addr
+
+    def perform_test(self, diagnostics_report):
+        resp = os.popen(
```
vpetersson, line 43: I have don't like `popen` or `subprocess` and would like to see us avoid them to the greatest degree possible. I know it gets a bit more complicated to use native features like `open()` instead, but I think it will make the code a lot easier to debug.

hw_diag/diagnostics/diagnostic.py
```diff
@@ -0,0 +1,15 @@
+
+class Diagnostic():
+    def __init__(self, key, friendly_name):
+        """
+        key - Key of relevant value in diagnostics_report dictionary
+        friendly_name - Same as key but a human_friendly_snake_case
```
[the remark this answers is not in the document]
vpetersson: 100% this. Also I created [this ticket](https://github.com/NebraLtd/hm-diag/issues/202) to get rid of the diagnostics dict.

marvinmarnold:
@kashifpk can you rebase this onto master? The diff currently looks weird and includes unrelated work.

kashifpk (approved):

hw_diag/diagnostics/bt_lte_diagnostic.py
kashifpk: @vpetersson - you mean to read  /sys/bus/usb/devices/*/idVendor file as regular file using open() and then do the grepping in python code?

I can handle that once hm-config is setup on my pi. Balena cloud is timing out.

How do I test the change? test cases would be mocked? So what is the acceptance criteria for this?

marvinmarnold:
The Compile Docker stage was hanging after 2h. I am going to manually restart.

hw_diag/diagnostics/key_diagnostics.py
```diff
@@ -0,0 +1,47 @@
...
+            return KeyDiagnostic(
+              key_mapping['key'],
+              key_mapping['friendly_name'],
+              key_mapping['key_path'])
+
+        self.key_diagnostics = map(get_diagnostic_for_key,
+                                   KEY_MAPPINGS)
+
+    def perform_test(self, diagnostics_report):
+        for key_diagnostic in self.key_diagnostics:
```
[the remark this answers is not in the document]
marvinmarnold: Not resolved. Punting: https://github.com/NebraLtd/hm-diag/issues/210

Issue #204, opened 2021-11-03
Ensure all exceptions thrown by miner are caught


marvinmarnold:
Related to: https://github.com/NebraLtd/hm-diag/issues/181
Exceptions created in this PR: https://github.com/NebraLtd/hm-pyhelper/pull/45

vpetersson:
Acceptance criteria:

- [ ] Stop all services other than hm-diag on the miner
- [ ] Make sure both the diagnostics page and the `/json` payload is being served with all other services stopped

marvinmarnold:
@SebastianMaj can you make sure [this invocation](https://github.com/NebraLtd/hm-diag/blob/master/hw_diag/tasks.py#L42) is also caught?

In testing, I am seeing:

```
 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1515, in full_dispatch_request
10.11.21 19:26:01 (-0600)  diagnostics      rv = self.handle_user_exception(e)
10.11.21 19:26:01 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1513, in full_dispatch_request
10.11.21 19:26:01 (-0600)  diagnostics      rv = self.dispatch_request()
10.11.21 19:26:01 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1499, in dispatch_request
10.11.21 19:26:01 (-0600)  diagnostics      return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
10.11.21 19:26:01 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/flask_caching/__init__.py", line 475, in decorated_function
10.11.21 19:26:01 (-0600)  diagnostics      rv = f(*args, **kwargs)
10.11.21 19:26:01 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/hw_diag-1.0-py3.10.egg/hw_diag/views/diagnostics.py", line 44, in get_diagnostics_json
10.11.21 19:26:01 (-0600)  diagnostics      diagnostics = read_diagnostics_file()
10.11.21 19:26:01 (-0600)  diagnostics    File "/usr/local/lib/python3.10/site-packages/hw_diag-1.0-py3.10.egg/hw_diag/views/diagnostics.py", line 32, in read_diagnostics_file
[...]

PR #66, opened 2021-06-21
Fix CI errors in diagnostics page redesign.

robputt:
Fixed CI errors in diagnostics page redesign.

PR #212, opened 2021-11-12
fix: Ensure miner exceptions are caught

SebastianMaj:
**Why**
Ensure all exceptions thrown by miner are caught and diagnostics page + /json payload is still served even upon exception

**How**
Added exception catching that works with hm-pyhelper. Caught many different exceptions being thrown in different parts of the repository. Stopped all other containers on the miner other than hm-diag and ensured that the diagnostics page and /json payload was still being served even if an exception was to occur.

**References**
[Issue #204](https://github.com/NebraLtd/hm-diag/issues/204)
[Issue #181](https://github.com/NebraLtd/hm-diag/issues/181)


Dockerfile
```diff
@@ -14,7 +14,8 @@ HEALTHCHECK \
 RUN \
     install_packages \
       i2c-tools \
-      usbutils
+      usbutils \
+      git
```
vpetersson, line 18: Why do we need `git` here?

shawaj: Also is it worth splitting this up into a build / run setup? Might save a bit of space also

SebastianMaj: > Why do we need `git` here?

I'll remove git from there, I was using it for testing hm-pyhelper from my subbranch without having to push to PyPi.

hw_diag/diagnostics/ecc_diagnostic.py
```diff
@@ -1,6 +1,7 @@
 import json
 from hm_pyhelper.miner_param import get_gateway_mfr_test_result
 from hm_pyhelper.diagnostics.diagnostic import Diagnostic
+from hw_diag.diagnostics.exceptions import EccMalfunctionException
```
marvinmarnold, line 4: Should import the original exception: https://github.com/NebraLtd/hm-pyhelper/blob/master/hm_pyhelper/exceptions.py#L9

hw_diag/diagnostics/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class EccMalfunctionException(Exception):
```
marvinmarnold, line 1: No need to duplicate this class.

hw_diag/tests/diagnostics/test_diagnostic_report.py
```diff
@@ -0,0 +1,46 @@
+import unittest
```
marvinmarnold, line 1: DiagnosticsReport was moved to hm_pythelper along with this [test](https://github.com/NebraLtd/hm-pyhelper/blob/master/hm_pyhelper/tests/test_diagnostic_report.py).

setup.py
```diff
@@ -17,7 +17,7 @@
 setup(
     name=base_name,
     version='1.0',
-    author=u'Nebra Ltd.',
+    author='Nebra Ltd.',
```
marvinmarnold, line 20: Any particular reason this was removed? Seems like its a Python 2 compatibility thing and not relevant either way in Python 3.

hw_diag/diagnostics/ecc_diagnostic.py
```diff
@@ -21,5 +22,5 @@ def perform_test(self, diagnostics_report):
                       str(json.dumps(ecc_tests))
                 diagnostics_report.record_failure(msg, self)
 
-        except Exception as e:
+        except EccMalfunctionException as e:
```
marvinmarnold, line 25: You are missing some other possible execptions:
- [GatewayMFRFileNotFoundException](https://github.com/NebraLtd/hm-pyhelper/blob/a98ff5e135fda14276239f66e6545d857af41362/hm_pyhelper/miner_param.py#L51)
- [ResourceBusyError](https://github.com/NebraLtd/hm-pyhelper/blob/a98ff5e135fda14276239f66e6545d857af41362/hm_pyhelper/lock_singleton.py#L66)

And [KeyDiagnostics](https://github.com/NebraLtd/hm-diag/blob/12420cfb08880a0b296b4130174f8563c8074ea1/hw_diag/diagnostics/key_diagnostics.py#L27) is an additional class that will need these checks. Should add unit tests for handling each exception.

hw_diag/utilities/miner.py
```diff
@@ -8,9 +9,21 @@ def fetch_miner_data(diagnostics):
     try:
         peerbook = client.get_peer_book()[0]
         height = client.get_height()
+        # Used for unit test to make sure exception
+        # can be thrown from hm-pyhelper
+        if "MC" in diagnostics and diagnostics['MC'] == 'testexception':
```
marvinmarnold, line 14: Is this code intentionally left in?

hw_diag/utilities/miner.py
```diff
@@ -8,9 +9,21 @@ def fetch_miner_data(diagnostics):
...
+            # Raise exception unit test error
+            raise MinerFailedFetchData("Unable to fetch miner data "
+                                       "from miner container "
+                                       "via JSON RPC API. Exception: %s" %
+                                       "Exception Test Error")
     except Exception as err:
-        raise Exception("Unable to fetch miner peerbook from miner container "
-                        "via JSON RPC API. Exception: %s" % str(err))
+        # Raise exception as MinerFailedFetchData exception from hm-pyhelper
+        raise MinerFailedFetchData("Unable to fetch miner data "
```
marvinmarnold, line 22: This change needs to be paired with some logic in [tasks.py](https://github.com/NebraLtd/hm-diag/blob/12420cfb08880a0b296b4130174f8563c8074ea1/hw_diag/tasks.py#L42) so that the exception is handled and doesn't prevent a report from being generated.

marvinmarnold (changes requested):

marvinmarnold:
Can you rebase on master? A lot of the commits in this PR are unrelated to the change and/or part of work that has already gotten merged in.

setup.py
SebastianMaj: Python 2 requires the author string to be in unicode? I found it odd that it was being converted from a byte string into a unicode one here, if this causes issues with python2 I will revert it immediately

hw_diag/diagnostics/ecc_diagnostic.py
SebastianMaj: Must've been from before, I changed this to be pulling from hm-pyhelper instead

hw_diag/utilities/miner.py
SebastianMaj: This is a unit test to make sure the exception can be thrown in the first place and that a valid connection to hm-pyhelper exists. diagnositcs['MC'] will = 'test exception' only when called from the unit test

SebastianMaj: Removed it as I was using side effect

SebastianMaj:
> Can you rebase on master? A lot of the commits in this PR are unrelated to the change and/or part of work that has already gotten merged in.

@marvinmarnold I rebased on master and force pushed the changes, should be reflected now. Just finalizing some changes and fixing a unit test before moving forward with a update push

SebastianMaj:
Unit test failing as it's relying on hm-pyhelper changes to exception formatting

hw_diag/diagnostics/key_diagnostics.py
```diff
@@ -25,8 +24,12 @@ def perform_test(self, diagnostics_report):
         # Try to get key, but there may be ECC lock or other failure
         try:
             public_keys = get_public_keys_rust()
+        except UnboundLocalError as e:
```
marvinmarnold, line 27: Is there a reason that the exceptions that we are catching here are different than the ones for `EccDiagnostic`? Both `get_gateway_mfr_result` and `get_public_keys_rust` invoke `run_gateway_mfr` so they should have the same exceptions.

It doesn't really matter since there is a catch-all `except Exception as e` in both cases.

marvinmarnold (changes requested):

hw_diag/diagnostics/ecc_diagnostic.py
```diff
@@ -21,5 +24,18 @@ def perform_test(self, diagnostics_report):
                       str(json.dumps(ecc_tests))
                 diagnostics_report.record_failure(msg, self)
 
+        except ECCMalfunctionException as e:
+            LOGGER.exception(e)
+            diagnostics_report.record_failure(str(e), self)
```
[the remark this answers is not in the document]
SebastianMaj: Oh, that's awesome! I'll remove the str conversion and just pass the exception directly. Thank you for letting me know 😊 !

hw_diag/diagnostics/key_diagnostics.py
SebastianMaj: Yeah so the reason this one has UnboundLocal is that I experienced it directly when testing this, I could add it to the rest but like you said we are catching all exceptions just in case one falls through that we didn't put a specific catch case for. In this specific scenario, I will add UnboundLocal to all of them it's just that since get_public_keys is relying on the rest there is a chance that if something fails it won't be bound so in this specific try catch you marked above it makes sense to include it to prevent running the rest in case it's unbound.

marvinmarnold: To be clear, I'm just talking about UnboundLocal. `get_public_keys_rust` can also potentially throw ECCMalfunctionException,  GatewayMFRFileNotFoundException, and ResourceBusyError. I'm also not sure it can still throw FileNotFoundError, NotADirectoryError as those are [now being caught](https://github.com/NebraLtd/hm-pyhelper/blob/95676fa2c05b8b8d7cb307313f91eb3353f7a653/hm_pyhelper/miner_param.py#L51) in hm-pyhelper.

SebastianMaj: > To be clear, I'm not just talking about UnboundLocal. `get_public_keys_rust` can also potentially throw ECCMalfunctionException, GatewayMFRFileNotFoundException, and ResourceBusyError. I'm also not sure it can still throw FileNotFoundError, NotADirectoryError as those are [now being caught](https://github.com/NebraLtd/hm-pyhelper/blob/95676fa2c05b8b8d7cb307313f91eb3353f7a653/hm_pyhelper/miner_param.py#L51) in hm-pyhelper.

Oh ok, I'll take a look at it and see what's being caught and what I can add, I just did a push with some changes let me know if they make sense. I'm currently looking at the changes and reviewing to see if Unit Tests can be added and I'll take a look at the functions you mentioned above and see what else I can catch

hw_diag/utilities/miner.py
```diff
@@ -8,9 +10,12 @@ def fetch_miner_data(diagnostics):
     try:
         peerbook = client.get_peer_book()[0]
         height = client.get_height()
+    except MinerFailedFetchData as err:
+        LOGGER.exception(err)
     except Exception as err:
-        raise Exception("Unable to fetch miner peerbook from miner container "
-                        "via JSON RPC API. Exception: %s" % str(err))
+        LOGGER.exception(err)
+        raise MinerFailedFetchData(err)\
```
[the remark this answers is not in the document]
SebastianMaj: I am now catching MinerFailedFetchData in tasks.py and setting 2 diagnostics values if they are not set, please review to see if this makes sense

hw_diag/tasks.py
```diff
@@ -39,7 +40,20 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
     diagnostics['AN'] = public_keys['name']
 
     # Fetch data from miner container.
-    diagnostics = fetch_miner_data(diagnostics)
+    try:
+        diagnostics = fetch_miner_data(diagnostics)
+    except MinerFailedFetchData as e:
+        if 'MC' not in diagnostics or ('MC' in diagnostics and diagnostics['MC'] <= 0 or diagnostics['MC'] is None):
```
marvinmarnold, line 46: There are two things that can be improved about this pattern:

1. The `if` statements should be removed. Even if previous values for these exist, we should record the failure when it happens.
2. We need to set values for more keys than just `MC` and `MN`. There are [other variables ](https://github.com/NebraLtd/hm-diag/pull/212/files#diff-57790e52a73cdf1e8700d99bf6714a8bb8cfbd3268f66bc652d8a74bba9b581aL15)that are normally set too.

```
    diagnostics['MC'] = peerbook.get('connection_count') > 1
    diagnostics['MD'] = peerbook.get('listen_addr_count') > 0
    diagnostics['MH'] = height.get('height')
    diagnostics['MN'] = peerbook.get('nat')
    diagnostics['MR'] = diagnostics['MN'] == 'symmetric'
```

marvinmarnold (changes requested):
Almost there. Can you also bump the pyhelper version and get the build passing before the next review?

hw_diag/tasks.py
SebastianMaj: Perfect, I'll set all the values to the error message. Would it make more sense to set the values to the simple message I currently have or set it to the exception itself

hw_diag/utilities/miner.py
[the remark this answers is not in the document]
marvinmarnold: With the code as-is, there's actually a downstream implication for these values. For example, `MH` is [later used](https://github.com/NebraLtd/hm-diag/blob/12420cfb08880a0b296b4130174f8563c8074ea1/hw_diag/tasks.py#L54) to determine the value for MS. So setting MH to -1 would make more sense than as an error message. After looking at the code, I think MH is the only field where we need to be careful.

hw_diag/tasks.py
marvinmarnold: With the code as-is, there's actually a downstream implication for these values. For example, MH is later used to determine the value for MS. So setting MH to -1 would make more sense than as an error message. After looking at the code, I think MH is the only field where we need to be careful.

SebastianMaj: I pushed the changes for this @marvinmarnold 

Would it make sense to set MC, MD, and MR to false as they seem to be conditionals or should a string with the error message be alright here, from what I've seen it should be fine as string here since MR will be set to false automatically since MN is the error string.

hw_diag/diagnostics/key_diagnostics.py
SebastianMaj: I'm now catching these in KeyDiagnostics the rest of it is all in hm_pyhelper the following functions are present there
`get_gateway_mfr_result` `get_public_keys_rust``run_gateway_mfr`

hw_diag/diagnostics/key_diagnostics.py
```diff
@@ -24,12 +26,18 @@ def perform_test(self, diagnostics_report):
...
+            diagnostics_report.record_failure(e, self)
         except (FileNotFoundError, NotADirectoryError) as e:
-            diagnostics_report.record_failure(str(e), self)
+            diagnostics_report.record_failure(e, self)
+        except ResourceBusyError as e:
+            diagnostics_report.record_failure(e, self)
         except Exception as e:
-            diagnostics_report.record_failure(str(e), self)
+            diagnostics_report.record_failure(e, self)
+            raise ECCMalfunctionException(e)\
```
marvinmarnold, line 39: We never want a diagnostic to throw an exception. This will cause the /initFile.txt endpoint to fail when we want it to be successful but return a payload that indicates some specific component is failing.

hw_diag/tasks.py
```diff
@@ -43,17 +43,21 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
...
-            diagnostics['MC'] = "Failed to Fetch"
-        if 'MN' not in diagnostics or ('MN' in diagnostics and diagnostics['MN'] is None):
-            diagnostics['MN'] = "Failed to Fetch"
+        diagnostics['MC'] = "Failed to Fetch"
+        diagnostics['MD'] = "Failed to Fetch"
+        diagnostics['MH'] = -1
+        diagnostics['MN'] = "Failed to Fetch"
+        diagnostics['MR'] = "Failed to Fetch"
         log.exception(e)
+        raise MinerFailedFetchData(str(e))\
```
marvinmarnold, line 59: We don't want to throw an exception here.

hw_diag/utilities/miner.py
```diff
@@ -8,9 +10,12 @@ def fetch_miner_data(diagnostics):
     try:
         peerbook = client.get_peer_book()[0]
         height = client.get_height()
+    except MinerFailedFetchData as err:
+        LOGGER.exception(err)
```
marvinmarnold, line 14: With this change, `MinerFailedFetchData` is being swallowed. tasks.py won't know if this error happens and won't be able to update the keys like MC, MD, etc correctly.

hw_diag/utilities/miner.py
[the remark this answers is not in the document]
marvinmarnold: That works. I would have probably set those values here and not re-raised an exception, but both work.

marvinmarnold (changes requested):

hw_diag/tasks.py
marvinmarnold: Good thought, I'm not sure. @ilyastrodubtsev will these values have any impact on what gets displayed on the dashboard?

hw_diag/utilities/miner.py
SebastianMaj: Thank you for catching this, with the tree of exceptions it gets confusing sometimes 😂

marvinmarnold (approved):
This looks good to go now.

hw_diag/tasks.py
```diff
@@ -39,7 +40,12 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
     diagnostics['AN'] = public_keys['name']
 
     # Fetch data from miner container.
-    diagnostics = fetch_miner_data(diagnostics)
+    try:
+        diagnostics = fetch_miner_data(diagnostics)
+    except MinerFailedFetchData as e:
```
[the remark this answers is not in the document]
SebastianMaj: I was thinking about leaving it just in case we need it in the future, I know it won't be called since it's no longer being raised in `fetch_miner_data` but you're right, it's better to remove it rather than leave unwanted code that may cause confusion.

SebastianMaj:
> Almost there. Can you also bump the pyhelper version and get the build passing before the next review?

Fixed one build issue, the other one is relying on an updated version of hm-pyhelper. [PR #66](https://github.com/NebraLtd/hm-pyhelper/pull/66) Last PR to remove all the other bump2version references and then do a full release and then I'll do another commit here on hm-diag to update the version of hm-pyhelper release and then build should pass.

badaczewski:
This is not fixed, and I'm a bit insulted to be dismissed to quickly. I was runnign the 11.17 firmware before this started happening.
image

PR #213, opened 2021-11-13
Adds dockerignore file to save space

vpetersson:
* Adds `.dockerignore` to avoid copying in `.git` into containers.


shawaj (approved):

shawaj:
@vpetersson do we need to add this for UPnP, miner, pktfwd as well?

shawaj:
Turns out this isn't actually needed for .git on balena...

https://www.balena.io/docs/learn/develop/dockerfile/#using-dockerfiles-with-balena

Screenshot_20211121-183009_Chrome

vpetersson:
Presumably this is only true if you're building the images on Balena. We're not. We're building using `docker buildx` and are then pushing to Docker Hub. These images are then consumed by Balena. Hence `.git` is already part of the image (unless excluded) by the time it is pulled down by Balena.

Issue #216, opened 2021-11-16
Add Helium API timeout


SebastianMaj:
Helium API on hm-diag is timing out but has a long period of timeout completely freezing up the dashboard from serving.

PR #217, opened 2021-11-16
Fix: Add timeout to request on helium api

SebastianMaj:
**Why**
Helium API is timing out, stuck on connecting to API while it's down

**How**
Added a timeout to the request and the dashboard served perfectly after this and displayed the timeout error

**References**
[Issue](https://github.com/NebraLtd/hm-diag/issues/216)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     build: .
     environment:
       - 'FIRMWARE_VERSION=2021.08.02.0'
-      - 'DEFAULT_TIMEOUT=5'
+      - 'HELIUM_API_TIMEOUT_SECONDS=5'
     volumes:
       - 'pktfwdr:/var/pktfwd'
       - 'miner-storage:/var/data'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     build: .
     environment:
       - 'FIRMWARE_VERSION=2021.08.02.0'
+      - 'DEFAULT_TIMEOUT=5'
     volumes:
       - 'pktfwdr:/var/pktfwd'
       - 'miner-storage:/var/data'
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -49,7 +49,7 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         log.exception(e)
 
     # Get the blockchain height from the Helium API
-    value = "1"
+    value = "Not Available"
     try:
         value = get_helium_blockchain_height()
     except KeyError as e:
@@ -69,6 +69,9 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
     diag_mh = int(diagnostics['MH'])
     diag_bch = int(diagnostics['BCH'])
     diagnostics['BSP'] = round(diag_mh / diag_bch * 100, 3)
+    
+    if value == "Not Available":
+        diagnostics['BSP'] = "Not Available"
 
     set_diagnostics_bt_lte(diagnostics)
 
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -49,14 +49,14 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         log.exception(e)
 
     # Get the blockchain height from the Helium API
-    value = "Not Available"
+    value = "1"
     try:
         value = get_helium_blockchain_height()
     except KeyError as e:
         logging.warning(e)
-    except (ConnectTimeout, ReadTimeout) as e:
+    except (ConnectTimeout, ReadTimeout):
         err_str = ("Request to Helium API timed out."
-            " Will fallback to block height of 1.")
+                   " Will fallback to block height of 1.")
         logging.exception(err_str)
     diagnostics['BCH'] = value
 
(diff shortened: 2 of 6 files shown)


hw_diag/utilities/blockchain.py
```diff
@@ -9,13 +9,18 @@ def get_helium_blockchain_height():
...
-        try:
-            result = result['data']['height']
-        except KeyError:
-            raise KeyError(
-                "Not found value from key ['data']['height'] in json"
-            )
-        return result
+    try:
+        result = requests.get('https://api.helium.io/v1/blocks/height',
+                              timeout=5)
```
marvinmarnold, line 14: Lets use an environment variable with a default value for the timeout.

hw_diag/utilities/blockchain.py
```diff
@@ -9,13 +9,18 @@ def get_helium_blockchain_height():
...
+        if result.status_code == 200:
+            result = result.json()
+            try:
+                result = result['data']['height']
+            except KeyError:
+                raise KeyError(
+                    "Not found value from key ['data']['height'] in json"
+                )
+            return result
+    except (requests.exceptions.ConnectTimeout,
```
marvinmarnold, line 24: Why bother catching these exceptions if we just rethrow them? Might as well remove the outter try/except to simplify the code in that case.

hw_diag/tasks.py
```diff
@@ -53,6 +54,8 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         value = get_helium_blockchain_height()
     except KeyError as e:
         logging.warning(e)
+    except (ConnectTimeout, ReadTimeout) as e:
+        logging.exception(e)
```
marvinmarnold, line 58: `logging.exception` will automatically show the stacktrace. So would be more useful to do something like `logging.exception("Request to Helium API timed out. Will fallback to block height of 1.")`.

marvinmarnold (changes requested):
Some small tweaks. Good tests.

hw_diag/utilities/blockchain.py
SebastianMaj: I removed the outer try/except in the code here, I originally was handling it in blockchain.py but it was much nicer to handle it outside in tasks

hw_diag/tasks.py
SebastianMaj: This makes sense, that way it'll give us a more detailed message of what is going on. I'll add that in

hw_diag/tasks.py
```diff
@@ -48,11 +49,15 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         log.exception(e)
 
     # Get the blockchain height from the Helium API
-    value = "1"
+    value = "Not Available"
```
marvinmarnold, line 52: Can you make sure these changes are compatible with the diagnostics [uploads schema](https://github.com/NebraLtd/hm-diag/blob/adc65aaed9b898313a9259c9ec0af9b566f228cf/bigquery/bq_schema.json#L111).

hw_diag/utilities/blockchain.py
```diff
@@ -11,7 +11,7 @@ def get_helium_blockchain_height():
     TypeError - if the key ['data']['height'] in response is not found.
     """
     result = requests.get('https://api.helium.io/v1/blocks/height',
-                            timeout=os.environ['DEFAULT_TIMEOUT'])
+                          timeout=os.environ['DEFAULT_TIMEOUT'])
```
marvinmarnold, line 14: This will cause an error if there's no environment variable set. We want it to use a hardcoded default. Like [this](https://github.com/NebraLtd/hm-config/blob/16f925719060275ba7ebc39ac7d676ab99743b26/gatewayconfig/__main__.py#L25).

```
timeout=os.getenv('HELIUM_API_TIMEOUT_SECONDS', 5)
```

hw_diag/tasks.py
SebastianMaj: This is not compatible, I'll set it back to 1. I'll have to see how the other values are affected referring to the big query schema and I may have to make changes to the view to accommodate for displaying a string instead

hw_diag/tests/test_blockchain_helium_height.py
```diff
@@ -18,6 +20,8 @@ class TestHelium(unittest.TestCase):
     the_response.json.return_value = get_data
     the_response.status_code = 200
 
+    os.environ['DEFAULT_TIMEOUT'] = '5'
```
marvinmarnold, line 23: Can this be removed?

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
     build: .
     environment:
       - 'FIRMWARE_VERSION=2021.08.02.0'
+      - 'HELIUM_API_TIMEOUT_SECONDS=5'
```
marvinmarnold, line 9: @vpetersson I've been exposing a lot of environment variables in services but haven't added them to docker-compose because they have defaults. But I like Sebastian's decision to add it here too. Self documenting. Do you think it makes sense to add all envvars here moving forward?

marvinmarnold (approved):

hw_diag/tests/test_blockchain_helium_height.py
SebastianMaj: Yeah that's true it's no longer needed, I have removed it

docker-compose.yml
SebastianMaj: I feel like this is a great way to expose the environment variables and allow them to be easily viewed/changed from build to build. Usually, when I deal with environment variables I either create a .env file as seen [here](https://docs.docker.com/compose/env-file/) or I include them in the docker-compose file as we are doing so now.

Issue #206, opened 2021-11-05
Change "RPI" key to Serial Number and add conditional checks


vpetersson:
We currently only support Raspberry Pi serial number lookup. As we add support for the RockPi, we need to make this more generic.

This task requires that we:

- [x] Rename the `RPI` key to `serial_number` 
- [x] Clean up the [Raspberry Pi serial number check](https://github.com/NebraLtd/hm-diag/blob/master/hw_diag/utilities/hardware.py#L57-L74)
- [x] Add support for reading the RockPi serial number (unclear how to do this at the moment)
- [x] Add a conditional serial number lookup that pulls in either the RockPi serial number or the Raspberry Pi serial number to populate the `serial_number` key

PR #219, opened 2021-11-17
Convert RPI to Serial Number

SebastianMaj:
**Why**
Need to convert RPI to Serial Number, read in RockPi serial number, and distinguish between RockPi and Raspberry Pi

**How**
- Replace any references of RPI to serial_number
- Fetch RockPi serial number from file (/proc/device-tree/serial-number) 
- @ilyastrodubtsev helped me discover this file and after looking deeper into it together we realized that this file also exists on the Raspberry Pi and contains just the serial number which means we can avoid the parsing we do with /proc/cpuinfo
- This removes the need to distinguish between RockPi and RaspberryPi as they both have this file in common

**References**
Issue #206

### DIFF
--- a/bigquery/bq_schema.json
+++ b/bigquery/bq_schema.json
@@ -52,7 +52,7 @@
   },
   {
     "mode": "REQUIRED",
-    "name": "RPI",
+    "name": "serial_number",
     "type": "STRING"
   },
   {
--- a/hw_diag/diagnostics/serial_number_diagnostic.py
+++ b/hw_diag/diagnostics/serial_number_diagnostic.py
@@ -1,20 +1,20 @@
 
 from hm_pyhelper.diagnostics.diagnostic import Diagnostic
 
-KEY = 'RPI'
+KEY = 'serial_number'
 FRIENDLY_NAME = "serial_number"
 SERIAL_FILEPATH = "/proc/device-tree/serial-number"
 
 
-class RpiSerialDiagnostic(Diagnostic):
+class SerialNumberDiagnostic(Diagnostic):
     def __init__(self):
-        super(RpiSerialDiagnostic, self). \
+        super(SerialNumberDiagnostic, self). \
             __init__(KEY, FRIENDLY_NAME)
 
     def perform_test(self, diagnostics_report):
         try:
-            rpi_serial = open(SERIAL_FILEPATH).readline()
-            diagnostics_report.record_result(rpi_serial, self)
+            serial_number = open(SERIAL_FILEPATH).readline()
+            diagnostics_report.record_result(serial_number, self)
 
         except FileNotFoundError as e:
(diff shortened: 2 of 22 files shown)


hw_diag/diagnostics/serial_number_diagnostic.py
```diff
@@ -1,9 +1,9 @@
...
 
-KEY = 'RPI'
-FRIENDLY_NAME = "raspberry_pi_serial_number"
-SERIAL_FILEPATH = "/proc/cpuinfo"
+KEY = 'serial_number'
+FRIENDLY_NAME = "serial_number"
+SERIAL_FILEPATH = "/proc/device-tree/serial-number"
 
 
 class RpiSerialDiagnostic(Diagnostic):
```
marvinmarnold, line 9: This class name should be updated too

marvinmarnold (changes requested):

bigquery/bq_schema.json
```diff
@@ -52,7 +52,7 @@
   },
   {
     "mode": "REQUIRED",
```
marvinmarnold, line 54: Also need to update `MH` to be nullable, right?

marvinmarnold (changes requested):
After this gets approved, please don't merge immediately. This will need to be coordinated with several other PRs.

bigquery/bq_schema.json
SebastianMaj: Yep I made MH, BCH, BSP all nullable, those are the main values that display numbers and hold weird numbers in the case of an error

SebastianMaj: Also no worries, I won't merge until given the ok to merge. Thank you for letting me know

marvinmarnold: > Yep I made MH, BCH, BSP all nullable, those are the main values that display numbers and hold weird numbers in the case of an error

My bad, didn't realize these were already nullable. Thanks.

marvinmarnold (dismissed):

marvinmarnold:
@SebastianMaj to get this unblocked by https://github.com/NebraLtd/hm-diag/issues/220, can you update the google cloud [upload logic](https://github.com/NebraLtd/hm-diag/blob/b1210ec87ab347d2a1ae8fe358b30fbf33bcb317/hw_diag/utilities/gcs_shipper.py#L39) so that it still uses `raspberry_pi_serial_number` for the upload?

marvinmarnold (approved):

shawaj:
> After this gets approved, please don't merge immediately. This will need to be coordinated with several other PRs.

Oops missed this comment sorry

vpetersson:
@shawaj Please don't merge other people's PRs going forward. We have a process for when and how we merge.

Issue #223, opened 2021-11-18
Improve error handling


shawaj:
Following on from #212 I'm not sure the way this is handled is particularly great.

For example, killing the miner container and then loading diagnostics shows this:
![Screenshot_20211118-052439_Chrome](https://user-images.githubusercontent.com/3359418/142357242-8c6058ae-02de-45ba-9af1-a18a7bfcbf94.jpg)

However, lots of the data can still be displayed, and there's no reason why the JSON should return an error for everything when it still can provide useful data and just errors on the unreachable info due to miner container being down for example.

We should just display all the data we can and only show error messages for individually missing info on those specific lines with specific error messages.

Cc @marvinmarnold @SebastianMaj

shawaj:
Relates to #204 and #207

KevinWassermann94:
I have removed some bits in the following with xxxx for security reasons, but you get the idea

{
  "diagnostics_passed": false,
  "errors": [
    "W0"
  ],
  "RPI": "00000000594xxxx",
  "raspberry_pi_serial_number": "00000000594xxxxx",
  "ECC": true,
  "E0": "00:BD:27:XX:XX:XX",
  "eth_mac_address": "00:BD:27:XX:XX:XX",
  "W0": "Failed to find filecontaining miner mac address.Exception: [Errno 2] No such file or directory: '/sys/class/net/wlan0/address'",
  "wifi_mac_address": "Failed to find filecontaining miner mac address.Exception: [Errno 2] No such file or directory: '/sys/class/net/wlan0/address'",
  "BN": "weathered-pond",
  "BALENA_DEVICE_NAME_AT_INIT": "weathered-pond",
  "ID": "ca4471543790929984d0627cbbd73c92",
  "BALENA_DEVICE_UUID": "ca4471543790929984d0627xxxxxx",
  "BA": "HELIUM-TESTNET",
  "BALENA_APP_NAME": "HELIUM-TESTNET",
  "FR": "868",
  "FREQ": "868",
  "FW": "2021.11.17.2-1",
  "FIRMWARE_VERSION": "2021.11.17.2-1",
  "VA": "NEBHNT-OUT1",
  "VARIANT": "NEBHNT-OUT1",
  "BT": true,
  "LTE": true,
  "LOR": true,
  "lora": true,
  "OK": "112osQb9FtajkbrLnahUiCPBQVu8",
  "onboarding_key": "112osQb9FtajkbrLna",
[...]

PR #229, opened 2021-11-18
fix: handle bad response for helium height

marvinmarnold:
**Why**
Diagnostics fail entirely if unable to get Helium blockchain height.

**How**
Return -1 if request fails.
It's not clear why the request is failing in the first place.
curl to the same endpoint works from miners that are actively failing.
Tried disabling cache on the request, but no impact.

**References**
Related to: https://github.com/NebraLtd/hm-diag/issues/223

### DIFF
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -48,8 +48,10 @@
                   <td>Sync Percentage</td>
                   {% if diagnostics.BSP > 99.98 %}
                     <td class="text-right">100%</td>
-                  {% elif diagnostics.BSP < 0 %}
+                  {% elif diagnostics.BSP == -100.00 %}
                     <td class="text-right">Not Available</td>
+                  {% elif not diagnostics.BSP %}
+                    <td class="text-right">API Not Available</td>
                   {% else %}
                     <td class="text-right">{{ '%0.2f' % diagnostics.BSP|float }}%</td>
                   {% endif %}
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -46,9 +46,9 @@
                 </tr>
                 <tr>
                   <td>Sync Percentage</td>
-                  {% if diagnostics.BSP > 99.98 %}
+                  {% if diagnostics.BSP and diagnostics.BSP > 99.98 %}
                     <td class="text-right">100%</td>
-                  {% elif diagnostics.BSP == -100.00 %}
+                  {% elif diagnostics.BSP and diagnostics.BSP == -100.00 %}
                     <td class="text-right">Not Available</td>
                   {% elif not diagnostics.BSP %}
                     <td class="text-right">API Not Available</td>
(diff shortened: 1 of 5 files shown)


KevinWassermann94:
I highly suspect that the Helium API is not reporting the blockchain height currently atm. A lot of things in the Explorer are not loading right now and I see the same behaviour on my SC units

SebastianMaj (approved):
Changes look good. Did we want to keep this as none to avoid having an issue when getting average or results from the backend?

shawaj:
Maybe we should grab the blockchain height from the miner over JSON RPC instead of from Helium API?

Or compare the two or something

hw_diag/templates/diagnostics_page.html
```diff
@@ -46,7 +46,9 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                 </tr>
                 <tr>
                   <td>Sync Percentage</td>
-                  {% if diagnostics.BSP > 99.98 %}
+                  {% if not diagnostics.BSP %}
+                    <td class="text-right">API Not Available</td>
```
vpetersson, line 50: "Helium API'

hw_diag/templates/diagnostics_page.html
```diff
@@ -58,6 +60,8 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                   <td>Height Status</td>
                   {% if diagnostics.MH == -1 %}
                     <td class="text-right">Not Available</td>
+                  {% elif not diagnostics.BCH %}
+                    <td class="text-right">API Not Available</td>
```
vpetersson, line 64: Helium API

KevinWassermann94:
Now seems to be a good time to try this out. Are we ready to merge to testnet?

hw_diag/templates/diagnostics_page.html
SebastianMaj: > "Helium API'

I changed it to say Helium API Not Available, will push soon

SebastianMaj:
> Now seems to be a good time to try this out. Are we ready to merge to testnet?

Should be good to merge soon, just testing this out locally to make sure it's all good. Balena is giving me trouble again with updating device but as soon as it goes through I'll run the tests

KevinWassermann94:
Yep Balena is having issues again.. Lot's of devices showing as Heartbeat only

SebastianMaj:
I've tested all the changes on my local dev miner, everything seems to be operational.

This is the diagnostics page with all containers running and API issues:
Screen Shot 2021-11-18 at 2 55 21 PM

This is the diagnostics page with the helium-miner container down and API issues:
Screen Shot 2021-11-18 at 2 49 19 PM

KevinWassermann94:
Let's start testing it then?

SebastianMaj:
> Let's start testing it then?

I'll try to get another code review on the changes just to be sure it's good to push, as I reviewed the changes Marvin had made but no one has reviewed the changes I pushed after

Issue #222, opened 2021-11-18
Possible regression: "all ok" when WiFi broken


shawaj:
@marvinmarnold I think in #221 we have introduced a regression perhaps.

Because PF is used for the "all ok" display on diagnostics page, it will now show "all ok" even when WiFi and LoRa are broken.

I don't think this was the intention of #221 - as that was supposed to be just for manufacturing

![Screenshot_20211118-021737_Chrome](https://user-images.githubusercontent.com/3359418/142339646-f41dd54a-26d8-4e6e-b693-b86af392cb70.jpg)

![Screenshot_20211118-021800_Chrome](https://user-images.githubusercontent.com/3359418/142339593-473c8018-e3b0-43f0-862c-58c81582c9b9.jpg)

shawaj:
Relates to https://github.com/NebraLtd/Hotspot-Production-Tool/pull/48

marvinmarnold:
I hadn't considered this, but knowing what I did at that time, I would have still done it this way. 

BUT, now that I understand HPT better, I see we could have easily created a new key that uses the custom logic in https://github.com/NebraLtd/hm-diag/pull/221 and only checked that from HPT. That way existing logic is unaffected.

shawaj:
Yeah it's not the end of the world, as the individual items still show their status correctly, but maybe can be fixed as part of fixing #223 perhaps just so the all ok call is "correct"

I also wonder if maybe @NebraLtd/tech-support have any other ideas

marvinmarnold:
[Consensus](https://nebraltd.slack.com/archives/C024BNQ1Y6T/p1637338884106000) is that we want to show OK when LOR, BT, E0, and ECC functional. But do not consider W0 as it is unreliable.

marvinmarnold:
After further discussion, W0 should still be used in manufacturing. "All ok" will be remove all together to avoid confusion: https://github.com/NebraLtd/product-management/issues/29

shawaj:
One thing with noting - PF is used in hm-config so if we remove PF it will cause issues in hm-config unless we rewrite that logic too

marvinmarnold:
Closing with follow-up: NebraLtd/product-management#29

PR #234, opened 2021-11-19
fix: LOR should be a factor for PF

marvinmarnold:
**Why**
We want to ignore W0 and only consider LOR, BT, E0, and ECC for determining PF.

**How**
Partial revert of: https://github.com/NebraLtd/hm-diag/pull/221/commits/205515f9a5e6aee6954e2f7ab11227b4305f2f17

**References**
- Related to: https://github.com/NebraLtd/hm-diag/issues/222


SebastianMaj (approved):
Looks good to me! 😊

Issue #232, opened 2021-11-19
Change Headers on API requests


lance36:
Hi, 

Now that the error-handling is solved I've noticed the Helium API dosent like python-requests without headers and is recently only responding with HTTP error `429`:
```
>>> import requests
>>> HELIUM_MINER_HEIGHT_URL = 'https://api.helium.io/v1/blocks/height'
>>> result = requests.get(HELIUM_MINER_HEIGHT_URL, timeout=5)
>>> result
<Response [429]>
```

I believe that is why the hotspots are showing an error right now, while the API is up and stable:
![image](https://user-images.githubusercontent.com/3243764/142648999-7ac02981-0178-4bc8-abaf-f3ad66acbf97.png)
![image](https://user-images.githubusercontent.com/3243764/142649050-279b6dd1-ec31-40d1-943d-e8335b201e6c.png)

Adding any different header to the request seems to solve the issue:
```
>>> headers = {'user-agent': 'Mozilla/5.0 (X11; U; Linux i686; it; rv:1.9.0.10) Gecko/2009042513 Ubuntu/8.04 (hardy) Firefox/3.0.10',}
>>> result = requests.get(HELIUM_MINER_HEIGHT_URL, headers=headers, timeout=5)
>>> result
<Response [200]>
>>> result.json()['data']['height']
1103908
```

Here: 
https://github.com/NebraLtd/hm-diag/blob/b1210ec87ab347d2a1ae8fe358b30fbf33bcb317/hw_diag/utilities/blockchain.py#L18-L19

vpetersson:
Nice find @lance36

marvinmarnold:
@SebastianMaj FYI we shouldn't close issues until they've reached production.

PR #236, opened 2021-11-19
fix: Helium API not returning

SebastianMaj:
**Why**
Helium API was not returning values because of missing headers 

**How**
Added headers specifying the user-agent field and it returned perfectly. Helium API is now happy with our request and returning a proper response

**References**
Closes: #232

### DIFF
--- a/hw_diag/utilities/blockchain.py
+++ b/hw_diag/utilities/blockchain.py
@@ -16,9 +16,10 @@ def get_helium_blockchain_height():
     TypeError - if the key ['data']['height'] in response is not found.
     """
     headers = {
-                ("user-agent': 'Mozilla/5.0 (X11; U;"
-                 " Linux i686; it; rv:1.9.0.10)"
-                 " Gecko/2009042513 Ubuntu/8.04 (hardy) Firefox/3.0.10"),
+                "user-agent": ("Mozilla/5.0 (X11; U;"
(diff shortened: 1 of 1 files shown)


lance36:
That should do the trick!
thanks @SebastianMaj

hw_diag/utilities/blockchain.py
```diff
@@ -15,7 +15,14 @@ def get_helium_blockchain_height():
     Possible exceptions:
     TypeError - if the key ['data']['height'] in response is not found.
     """
+    headers = {
+                "user-agent": ("Mozilla/5.0 (X11; U;"
+                               " Linux i686; it; rv:1.9.0.10)"
+                               " Gecko/2009042513 Ubuntu/8.04"
+                               " (hardy) Firefox/3.0.10"),
+              }
```
vpetersson, line 23: @SebastianMaj let's use a more modern string, like: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.1 Safari/605.1.15`

SebastianMaj: @vpetersson Done it's now updated, I'll push it to the miner and report results

vpetersson (approved):

Issue #237, opened 2021-11-20
RockPi support


shawaj:
We need to enable RockPi support ASAP.

Requires:
#96 
#88 

Also we don't seem to have GPIO support for RockPi currently.

Here is an example device:
https://dashboard.balena-cloud.com/devices/963415259c442625fc5f5ac241f48578/summary

Acceptance criteria
- Add documentation
- ECC issue: https://nebraltd.slack.com/archives/C024BNQ1Y6T/p1638294078420000

shawaj:
Related to https://github.com/NebraLtd/hm-config/issues/96

shawaj:
https://github.com/NebraLtd/hm-diag/blob/fd134a4e7d9211e9ff9bc0de9840df6bb8246874/hw_diag/utilities/hardware.py#L41-L54 needs fixing

shawaj:
https://github.com/NebraLtd/hm-pyhelper/blob/d70f3659e4ca67185c6d90e026d3ef5e06145c5c/hm_pyhelper/miner_param.py#L30-L32

this is currently outputting gateway_mfr test --path /dev/i2c-7 but it needs to be gateway_mfr --path /dev/i2c-7 test

shawaj:
Need to bump hm-pyhelper https://github.com/NebraLtd/hm-pyhelper/pull/68

shawaj:
Closed by https://github.com/NebraLtd/helium-miner-software/pull/286

PR #238, opened 2021-11-20
fix: support RockPi with i2c-7

shawaj:
**Why**

Add support for RockPi which uses i2c-7

**How**

Add i2c-7 to the check

**References**

#237


shawaj:
Also need to bump hm-pyhelper to bring in these changes https://github.com/NebraLtd/hm-pyhelper/pull/68

marvinmarnold (approved):

Issue #96, opened 2021-08-06
Bluetooth detection: better way to identify BT interface


shawaj:
Similar to https://github.com/NebraLtd/hm-diag/issues/88 but for BT instead of LTE

Currently, we identify "BT detected" by looking for the USB ID of the bluetooth module:

https://github.com/NebraLtd/hm-diag/blob/2a6efbaaa92d715c7462419ebd471b43da519d06/hw_diag/utilities/hardware.py#L8-L21

Whilst this works, it is a little bit cumbersome IMO, and doesn't allow checking for in-built bluetooth hardware (which we will have on future hardware - for example the light hotspots).

So ideally, we need a better way to identify the bluetooth interface. Perhaps by checking for a device with `hcitool dev` or `hciconfig -a`

shawaj:
@marvinmarnold @vpetersson  need to get this resolved in order for diagnostics to work properly with RockPi

marvinmarnold:
Will keep an eye on this. Its in our next sprint (started today). @posterzh is about to start working on it.

shawaj:
Sweet 👌

posterzh:
There are 3 utilities for us to get BLE devices. 
- `hcitool dev`
- `hciconfig -a`
- `bluetoothctl list`
![image](https://user-images.githubusercontent.com/91350131/142783435-210a4ef6-0bec-4b1b-8611-eb5527b6b025.png)

vpetersson:
Well, there are more ways. Those are just CLI tools for *interacting* with a BLE device. You can find out if a BLE device is connected on kernel level instead (such as looking in `/proc` or `/sys` for instance).

posterzh:
@vpetersson I implemented getting the BLE devices and LTE devices programatically using dbus.
https://github.com/NebraLtd/hm-diag/compare/master...posterzh/better-detect-bt#diff-1bc7190be6e1a8de2f5a4d659238a140f25dfff6cffa3587bc2e317b6d430709R57-R107

This draft implementation would fix both of https://github.com/NebraLtd/hm-diag/issues/96 and https://github.com/NebraLtd/hm-diag/issues/88
Also I confirm that it's possible to use all the features of `nmcli` cli using dbus.


To interact BLE and LTE devices with other than dbus, I played with several commands and they needs the `diagnostics` container to set `network_mode` to host and `NET_ADMIN` capability.
https://www.balena.io/docs/learn/develop/multicontainer/
[...]

Issue #88, opened 2021-07-18
LTE detected should support all tested compatible LTE modules


shawaj:
See list of tested units here:

https://github.com/NebraLtd/helium-miner-software/wiki/3G---4G-Module-Compatability-List

Only currently looks for 2c7c in USB vendor ID

https://github.com/NebraLtd/hm-diag/blob/1444a3b2ccfcec7c6b505ede8dd2ecfa31b1e831/diagnostics-program/main.py#L91

shawaj:
Perhaps it should instead use `mmcli -L` or similar to detect a working modem instead?

shawaj:
EG25G-S = ID 2c7c:0125 Quectel Wireless Solutions Co., Ltd. EC25 LTE modem
EG25G = ID 2c7c:0125 Quectel Wireless Solutions Co., Ltd. EC25 LTE modem
sim7100e = ID 1e0e:9001 Qualcomm / Option
reyax ryx9000 = ID 1bc7:1201 Telit Wireless Solutions
ec25-e = ID 2c7c:0125 Quectel Wireless Solutions Co., Ltd. EC25 LTE modem
eg21-g = ID 2c7c:0121 Quectel Wireless Solutions Co., Ltd. EC21 LTE modem
BG95-M3 = ID 2c7c:0700 Quectel Wireless Solutions Co., Ltd.
BG96 = ID 2c7c:0296 Quectel Wireless Solutions Co., Ltd. BG96 CAT-M1/NB-IoT modem
Sierra mc7700 = ID 1199:68a2 Sierra Wireless, Inc.

shawaj:
See also table at https://github.com/NebraLtd/Helium-Guides/blob/main/docs/outdoor-hotspot/lte-config.md

shawaj:
Huawei ME909s-120 = ID 12d1:15c1 Huawei Technologies Co., Ltd. ME906s LTE M.2 Module
```
root@c03ac42:~# mmcli -m 0
  -----------------------------
  General  |         dbus path: /org/freedesktop/ModemManager1/Modem/0
           |         device id: cbcec58bece9c9fc9b2fb5c338dbca65d2c3f715
  -----------------------------
  Hardware |      manufacturer: Huawei Technologies Co., Ltd.
           |             model: HUAWEI Mobile V7R11
           | firmware revision: 11.617.01.00.00
           |      h/w revision: RM1ME909ASM
           |         supported: gsm-umts, lte
[...]

PR #242, opened 2021-11-22
 Better detect BT devices and LTE devices

posterzh:
**Why**
Detect the BLE and LET devices in better way.

**How**
Get the BLE devices and LTE devices programatically using dbus.

**References**
https://github.com/NebraLtd/hm-diag/issues/96 
https://github.com/NebraLtd/hm-diag/issues/88

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,17 +10,20 @@ RUN mkdir /tmp/build
 COPY ./ /tmp/build
 WORKDIR /tmp/build
 
+ENV PATH="/root/.local/bin:$PATH"
+
 RUN \
     install_packages \
             python3-dev \
             python3-minimal \
             python3-pip \
+            python3-setuptools \
             libgirepository1.0-dev \
             gcc \
             pkg-config \
             libdbus-1-dev && \
-    pip install --no-cache-dir --user -r requirements.txt && \
-    pip install --no-cache-dir --user .
+    pip3 install --no-cache-dir --user -r requirements.txt && \
+    pip3 install --no-cache-dir --user .
 
 # No need to cleanup the builder
 
@@ -32,7 +35,7 @@ FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner
 RUN \
     install_packages \
         i2c-tools \
-        libdbus-1-dev \
+        libdbus-1-3
 
 HEALTHCHECK \
     --interval=120s \
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,9 +45,4 @@ COPY --from=builder "$PYTHON_DEPENDENCIES_DIR" "$PYTHON_DEPENDENCIES_DIR"
 ENV PYTHONPATH="${PYTHON_DEPENDENCIES_DIR}:${PYTHONPATH}"
 ENV PATH="${PYTHON_DEPENDENCIES_DIR}/bin:${PATH}"
 
-# Cleanup
-RUN apt-get autoremove -y && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
 ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:5000", "hw_diag:wsgi_app"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,11 @@
 # The balenalib/raspberry-pi-debian-python image was tested but missed many dependencies.
 FROM balenalib/raspberry-pi-debian:buster-build-20211014 as builder
 
+ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
+
 RUN mkdir /tmp/build
 COPY ./ /tmp/build
 WORKDIR /tmp/build
-
-ENV PATH="/root/.local/bin:$PATH"
 
 RUN \
     install_packages \
@@ -22,8 +22,8 @@ RUN \
             gcc \
             pkg-config \
             libdbus-1-dev && \
-    pip3 install --no-cache-dir --user -r requirements.txt && \
-    pip3 install --no-cache-dir --user .
+    pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" -r requirements.txt && \
+    pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
 
 # No need to cleanup the builder
 
@@ -31,6 +31,8 @@ RUN \
 ################################### Stage: runner ##################################################
 
 FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner
+
+ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
 
 RUN \
     install_packages \
@@ -44,8 +46,20 @@ HEALTHCHECK \
     --retries=10 \
   CMD wget -q -O - http://0.0.0.0:5000/initFile.txt || exit 1
 
-# Copy packages from builder and update PATH to activate it
-COPY --from=builder /root/.local /root/.local
-ENV PATH="/root/.local/bin:$PATH"
+# Copy packages from builder
+COPY --from=builder "$PYTHON_DEPENDENCIES_DIR" "$PYTHON_DEPENDENCIES_DIR"
 
-ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:5000", "hw_diag:wsgi_app"]
+# Add python dependencies to PYTHONPATH
+ENV PYTHONPATH="${PYTHON_DEPENDENCIES_DIR}:${PYTHONPATH}"
+ENV PATH="${PYTHON_DEPENDENCIES_DIR}/bin:${PATH}"
+
+# gunicorn depends on "/usr/bin/python3"
+RUN ln -s /usr/local/bin/python3 /usr/bin/python3
+
+# Cleanup
+RUN apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+#ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:5000", "hw_diag:wsgi_app"]
+ENTRYPOINT ["tail", "-f", "/dev/null"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,24 +10,17 @@ RUN mkdir /tmp/build
 COPY ./ /tmp/build
 WORKDIR /tmp/build
 
-# This will be the path that venv uses for installation below
-ENV PATH="/opt/venv/bin:$PATH"
-
 RUN \
     install_packages \
             python3-dev \
             python3-minimal \
             python3-pip \
-            python3-venv \
             libgirepository1.0-dev \
             gcc \
             pkg-config \
             libdbus-1-dev && \
-    # Because the PATH is already updated above, this command creates a new venv AND activates it
-    python3 -m venv /opt/venv && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,28 +1,74 @@
...
+            python3-dev=3.7.3-1 \
+            python3-minimal=3.7.3-1 \
+            python3-pip=18.1-5+rpt1 \
+            python3-venv=3.7.3-1 \
+            libgirepository1.0-dev=1.58.3-2 \
+            gcc=4:8.3.0-1+rpi2 \
+            pkg-config=0.29-6 \
+            libdbus-1-3 && \
+    # Because the PATH is already updated above, this command creates a new venv AND activates it
+    python3 -m venv /opt/venv && \
```
shawaj, line 32: IIRC @marvinmarnold and @vpetersson mentioned previously that there was little point in using venv in containerised setup.

Think was in relation to hm-config but guess it applies here too

Dockerfile
```diff
@@ -1,28 +1,74 @@
...
+####################################################################################################
+################################### Stage: runner ##################################################
+
+FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner
+
+# Install bluez, libdbus, network-manager, python3-gi, and venv
+RUN \
+    install_packages \
+        i2c-tools \
+        bluez \
```
shawaj, line 47: Do we need bluez and network manager in the container if we are communicating with them over dbus?

Dockerfile
```diff
@@ -1,28 +1,74 @@
...
+
+# This will be the path that venv uses for installation below
+ENV PATH="/opt/venv/bin:$PATH"
+
+# Install python3-minimal, pip3, wget, venv.
+# Then set venv environment copied from builder.
+# Finally, use pip to install dependencies.
+RUN \
+    install_packages \
+            python3-dev=3.7.3-1 \
```
shawaj, line 23: IIRC @vpetersson thought it best to unpin packages previously

Dockerfile
marvinmarnold: I still have a soft spot for venv since it makes python portable. But yea, we did agree to stop using it.

@posterzh hm-pktfwd is an example of the new pattern: https://github.com/NebraLtd/hm-pktfwd/blob/e067bef1be17f9d3c1036cbe1100b317d5cc5f57/Dockerfile#L65

Dockerfile
posterzh: Make sense

Dockerfile
posterzh: I will check.

vpetersson:
Nice improvement, @posterzh!

Dockerfile
posterzh: Nice pattern.

Dockerfile
posterzh: @shawaj Without the dependencies of `bluez`, `network-manager`, `modem-manager`, detecting BT, WIFI, LTE devices works well with only `dbus` dependency. I removed unnecessary dependencies in the Dockerfile.

posterzh:
I am going to migrate dbus python library from `dbus-python` to `pydbus`. 
- `dbus-python` is deprecated(https://github.com/posborne/dbus-python).
image
- To install `dbus-python`. we need to install `gcc` and `python-dev` package. multistage docker build is used for optimization for the size of container.
- `pydbus`(https://github.com/LEW21/pydbus) are installed well without above dependency in `raspberry-pi-debian-python` system.

Dockerfile
shawaj: @marvinmarnold I am not fussed either way! Whichever you think is best TBH

posterzh:
**Detecting BT and LTE devices using dbus:**

image

**Diagnostic Web page:**

image

**Diagnostic container from the multistage docker build:**

image

shawaj:
Looks good to me, but probably worth @vpetersson or @marvinmarnold or someone else from @NebraLtd/developers taking a deeper look.

vpetersson:
Good job @posterzh. This is a big improvement. I wonder however if we should move the detection logic to hm-pyhelper` instead as we should probably re-use this code in hm-diag too. Wdyt @marvinmarnold?

brianwaller063:
> Good job @posterzh. This is a big improvement. I wonder however if we should move the detection logic to hm-pyhelper` instead as we should probably re-use this code in hm-diag too. Wdyt @marvinmarnold?

FYI, in the `hm-diag`, the only feature used is to check the presence of BT/LTE devices. Furthermore, I added the utility methods to get the details of all BT, WiFi, LTE(`get_bt_devices()`, `get_wifi_devices()`, `get_lte_devices()`). These details of devices are used in checking the presence for validating their state. Meanwhile, almost the same code can be used for managing and connecting WiFi devices.

posterzh:
> Good job @posterzh. This is a big improvement. I wonder however if we should move the detection logic to hm-pyhelper` instead as we should probably re-use this code in hm-diag too. Wdyt @marvinmarnold?

FYI, in the `hm-diag`, the only feature used is to check the presence of BT/LTE devices. Furthermore, I added the utility methods to get the details of all BT, WiFi, LTE(`get_bt_devices()`, `get_wifi_devices()`, `get_lte_devices()`). These details of devices are used in checking the presence for validating their state. Meanwhile, almost the same code can be used for managing and connecting WiFi devices.

hw_diag/utilities/hardware.py
```diff
@@ -1,11 +1,70 @@
 import logging
-import os
 from time import sleep
-
+import dbus
 from hm_pyhelper.miner_param import get_public_keys_rust
 from hm_pyhelper.hardware_definitions import variant_definitions, is_rockpi
 from hw_diag.utilities.shell import config_search_param
 
+log = logging.getLogger()
```
marvinmarnold, line 8: Can you use get_logger from hm_pyhelper?

hw_diag/utilities/hardware.py
```diff
@@ -113,3 +259,7 @@ def get_public_keys_and_ignore_errors():
         }
 
     return public_keys
+
+
+if __name__ == '__main__':
+    print('set_diagnostics_bt_lte: %s' % set_diagnostics_bt_lte({}))
```
marvinmarnold, line 265: This should be log instead of print.

marvinmarnold (changes requested):
This work looks really good. The code is very readable and clear.

This logic also needs to be added to https://github.com/NebraLtd/hm-diag/blob/00c716dd4980e35ac7436d9f59aebb17bec07535/hw_diag/diagnostics/bt_lte_diagnostic.py Until we have completed https://github.com/NebraLtd/Hotspot-Production-Tool/issues/28 we will continue to have parallel logic for /json and /initFile.txt.

Lastly, my typical comment. Could you please add some basic unit tests?

hw_diag/utilities/hardware.py
posterzh: Ok, make sense.

vpetersson (changes requested):

hw_diag/utilities/hardware.py
posterzh: Ok

hw_diag/utilities/hardware.py
```diff
@@ -15,25 +74,112 @@ def should_display_lte(diagnostics):
...
+                    "Powered": str(adapter.get("Powered")),
+                    "Discoverable": str(adapter.get("Discoverable")),
+                    "Pairable": str(adapter.get("Pairable")),
+                    "Discovering": str(adapter.get("Discovering")),
+                })
+
+        log.info('BLE Devices: %s' % ble_devices)
+    except dbus.exceptions.DBusException as ex:
+        log.error(ex.get_dbus_message())
+    except Exception as ex:
```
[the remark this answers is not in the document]
posterzh: I agree. I will keep in mind it.

Dockerfile
```diff
@@ -11,18 +38,16 @@ HEALTHCHECK \
...
-WORKDIR /tmp/build
+# Add python dependencies to PYTHONPATH
+ENV PYTHONPATH="${PYTHON_DEPENDENCIES_DIR}:${PYTHONPATH}"
+ENV PATH="${PYTHON_DEPENDENCIES_DIR}/bin:${PATH}"
 
-RUN \
-    pip3 install --no-cache-dir -r /tmp/build/requirements.txt && \
-    python3 setup.py install && \
-    rm -rf /tmp/build
+# Cleanup
```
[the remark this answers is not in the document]
posterzh: Yes, the size of the docker image is same after removing the clean up.

vpetersson:
@posterzh Please fix the linting errors.

Dockerfile
```diff
@@ -1,8 +1,35 @@
...
+
+RUN mkdir /tmp/build
+COPY ./ /tmp/build
+WORKDIR /tmp/build
+
+RUN \
+    install_packages \
+            build-essential \
+            libdbus-glib-1-dev && \
+    pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
```
vpetersson, line 18: Nitpicking here, but I'd probably run this on a separate `RUN` line just to help with debugging if for whatever reason it would fail building.

vpetersson (approved):
Approved by me, but I would ideally like to see some unit tests here. Don't merge this before @marvinmarnold approves it.

Dockerfile
posterzh: It would help debugging.

posterzh: > Approved by me, but I would ideally like to see some unit tests here. Don't merge this before @marvinmarnold approves it.

Yeah, I am writing the unit test and a parallel logic in `DiagnosticsReport`. These pushes are just for `In-progress`.

vpetersson: I see. Please update the PR to 'Draft' mode if that is the case.

shawaj:
> Good job @posterzh. This is a big improvement. I wonder however if we should move the detection logic to hm-pyhelper` instead as we should probably re-use this code in hm-diag too. Wdyt @marvinmarnold?

@vpetersson do you think we would need this in hm-config? We seem to be using some custom nmcli in there so not sure it'll be compatible https://github.com/NebraLtd/hm-config/blob/3b2e0fe138ab04533ffa44afa3d0a60d7972430e/gatewayconfig/processors/wifi_processor.py#L20

posterzh:
> This work looks really good. The code is very readable and clear.
> 
> This logic also needs to be added to https://github.com/NebraLtd/hm-diag/blob/00c716dd4980e35ac7436d9f59aebb17bec07535/hw_diag/diagnostics/bt_lte_diagnostic.py Until we have completed [NebraLtd/Hotspot-Production-Tool#28](https://github.com/NebraLtd/Hotspot-Production-Tool/issues/28) we will continue to have parallel logic for /json and /initFile.txt.
> 
> Lastly, my typical comment. Could you please add some basic unit tests?

@marvinmarnold 
Unit tests are added.
What about using a separate `BtDiagnostics` and `LteDiagnostics` rather than `BtLteDiagnostics`?
https://github.com/NebraLtd/hm-diag/blob/c7253186e983d09e336f58fcefe5b69168e12c7d/hw_diag/diagnostics/bt_lte_diagnostic.py#L60

marvinmarnold (approved):
> What about using a separate BtDiagnostics and LteDiagnostics rather than BtLteDiagnostics?

Good idea but lets not let that hold things up. Merge this and create a follow up ticket  to do in a future sprint.

posterzh:
@marvinmarnold I created a follow up ticket as above.

Issue #248, opened 2021-11-25
Investigate slow response times from endpoints on some devices


kashifpk:
Slack discussion: https://nebraltd.slack.com/archives/C024BNQ1Y6T/p1637747430275400

TLDR;
-  `/version` endpoint taking somewhere around 1 to 7 minutes to respond.
- `/json` and `/initFile.txt` timing out or returning empty response.

This is only happening with gunicorn running on port 5000. Accessing these endpoints using flask's development server or by running gunicorn on a different port gives results back instantly.

kashifpk:
The connections to hm-diag seem to be in `FIN_WAIT2` state

![image](https://user-images.githubusercontent.com/198638/143373111-ccc0d93e-44d2-4361-903d-1dab00e2f335.png)

> FIN_WAIT_2 seems to occur when the server has an active connection with a client and wants to shut down the TCP connection (probably in response to a normal application layer "exit"). ... From the server's perspective, the connection is now closed, and the server can't send any more data. Since this is on multiple connections that probably takes up all gunicorn threads that are running causing the slow response times.

To check if this was something related to the client not closing the connection properly (unlikely), I tried setting kee-alive to false, and specifying timeouts but that didn't solve the issue.

![image](https://user-images.githubusercontent.com/198638/143373875-9f64fa0f-5bf8-4314-b570-2cc2bf3d101d.png)

kashifpk:
Apparently this is a known issue with docker on linux

I've dome some investigations into stuck connections in FIN-WAIT-2 state for our Nimbus Eth 2.0 Node project which we were running with Docker 20.10.2 on Ubuntu 20.04.1, and the solution that we've chosen was to ditch the docker-proxy by setting userland-proxy:false, which as far as I can tell has fixed the problem with the growing number of FIN-WAIT-2 connections.

https://github.com/docker/for-linux/issues/335

One solution provided was:

> In my opinion the issue is caused by the combination of docker-proxy and a few other circumstances. I say this because we have many other pieces of software using docker-proxy without such stuck connections, so it's not just docker-proxy that triggers this issue.

But need to see if `balena-engine` supports this or not and if this is something we can configure or it's baked into balena images.

This does not happen inside the container itself

```shell
# netstat -pnt | grep FIN_WAIT | wc -l
0
```

There is a more in-depth discussion here:

https://github.com/status-im/infra-nimbus/issues/35

But so far definitely seems like a docker bug.

kashifpk:
[...]

PR #249, opened 2021-11-25
Fix slow responses issue by switching to port 7000

kashifpk:
See the original discussion in #248 .

Basically using a different port instead of the current 5000  for gunicorn solves the issue.

Merging of this PR should be done in parallel with:  https://github.com/NebraLtd/helium-miner-software/pull/276

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,4 @@ RUN \
     python3 setup.py install && \
     rm -rf /tmp/build
 
-ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:5000", "hw_diag:wsgi_app"]
+ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:7000", "hw_diag:wsgi_app"]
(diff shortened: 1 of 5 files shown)


kashifpk:
Update on this. Moving my Pi to @marvinmarnold's dev-nebra fleet updated it with the image he had running on it (which was also from latest master) and there were not delays there. So we haven't been able to pin-point the cause of this issue. Also this issue isn't appearing on actual miner.

I'll keep monitoring testnet for this as we did just push `/version` endpiont related code to master so it should be in testnet. If this appears in testnet then we should investigate further. Otherwise if it's not happening on the miners then it's probably not worth the time for investigation.

vpetersson:
I don't want to merge this until we have seen some hard evidence for this (e.g. the tests) and/or understand *why* changing port fixes this issue.

kashifpk:
> I don't want to merge this until we have seen some hard evidence for this (e.g. the tests) and/or understand _why_ changing port fixes this issue.

I agree.  I'll convert it to draft for the time being and later either close it or make appropriate changes if we see this issue in testnet.

robputt:
I would suggest closing this, not sure it makes sense.

shawaj:
Quite an old one that's been abandoned and don't think we found this to be an issue in the end

kashifpk:
Yes we never experienced it on our production miners.

Issue #235, opened 2021-11-19
Improve diagnostics page messages when Helium API and miner RPC unavailable


marvinmarnold:
**User Story**
As a diagnostics user, I should be able to clearly differentiate miner errors from Helium API errors so that I have a better sense of what is going wrong.

**Notes**
- We were previously using negative values to indicate errors but have since started using `None` values.

[Slack convo](https://nebraltd.slack.com/archives/C0XGF69SN/p1637336138314900?thread_ts=1637334745.312300&cid=C0XGF69SN)

**Acceptance Criteria**
- Sync Percentage: Remove [Not Available](https://github.com/NebraLtd/hm-diag/blob/00c716dd4980e35ac7436d9f59aebb17bec07535/hw_diag/templates/diagnostics_page.html#L53) from template. This is now no longer possible (we use None instead of -1). 
- Sync Percentage: Add `Miner RPC Unavailable`
- Sync Percentage: Rename `Helium API Not Available` to `Helium API Unavailable`
- Height Status: Rename `Not Available` to `Miner RPC Unavailable` 
- Height Status: Add `Helium API Unavailable`.

shawaj:
Related to #195

PR #256, opened 2021-11-29
fix: display Miner RPC/Helium API availability correctly

posterzh:
**Why**
As a diagnostics user, I should be able to clearly differentiate miner errors from Helium API errors so that I have a better sense of what is going wrong.

**How**
- Update view template `diagnostics_page.html` to display the both of Miner RPC and Helium API status correctly.

**References**
https://github.com/NebraLtd/hm-diag/issues/235

### DIFF
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -46,22 +46,26 @@
                 </tr>
                 <tr>
                   <td>Sync Percentage</td>
-                  {% if not diagnostics.BSP %}
-                    <td class="text-right">Helium API Not Available</td>
+                  {% if not diagnostics.BCH and not diagnostics.MH %}
+                    <td class="text-right">Helium API and Miner RPC Unavailable</td>
+                  {% elif not diagnostics.BCH %}
+                    <td class="text-right">Helium API Unavailable</td>
+                  {% elif not diagnostics.MH %}
+                    <td class="text-right">Miner RPC Unavailable</td>
                   {% elif diagnostics.BSP > 99.98 %}
                     <td class="text-right">100%</td>
-                  {% elif diagnostics.BSP == -100.00 %}
-                    <td class="text-right">Not Available</td>
                   {% else %}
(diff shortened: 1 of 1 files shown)


posterzh:
@marvinmarnold  Do you know a fleet that can replicate an error when calling the helium miner api?

shawaj:
@posterzh you can replicate it in any fleet by just killing the miner container

posterzh:
> @posterzh you can replicate it in any fleet by just killing the miner container

I want to replicate the error in helium blockchain api call.
https://github.com/NebraLtd/hm-diag/blob/7e26f354a01476273b03766fe33b16dc92d943fa/hw_diag/utilities/blockchain.py#L10

Sometimes, calling `get_helium_blockchain_height()` failed, but CURL command succeed.
I saw that behavior once before, but it's hard to replicate it now.

shawaj:
@posterzh I don't think there is a reliable way to replicate that. However one possible is during a snapshot load when the miner is using a lot of resources... This normally causes the API to fail.

@robputt might have some other ideas

posterzh:
The subject was the REST API call to get the helium blockchain height. 
Marvin confirm that the API requests are no longer failing.

posterzh:
When diagnostics['BCH'] is None and diagnostics['MH'] is None:
image

When diagnostics['BCH'] is None:
image

When diagnostics['MH'] is None:
image

When all valid:
image

marvinmarnold:
> I want to replicate the error in helium blockchain api call.

I have a process for debugging this, but not its only useful if the API is actually returning 400. To do that, I SSH into a diagnostics container then run:

```
$ python3.10

>>> from hw_diag.utilities.blockchain import get_helium_blockchain_height
>>> get_helium_blockchain_height()
```

Depending on what mood the API is in, that request may or may not succeed. But even when that request fails, the corresponding curl request from the same diagnostics container will succeed: `curl https://api.helium.io/v1/blocks/height`

marvinmarnold (approved):

Issue #264, opened 2021-12-03
Manufacturing broken because of gateway_mfr error


kashifpk:
@shawaj @SebastianMaj @marvinmarnold 

Manufacturing is using images sent to them on Dec 2, 2021. FIRMWARE_VERSION=2021.11.30.1. We still see the error `gateway_mfr test finished with error`. This means that we're not on-boarding any units until this gets cleared.

This is the screenshot of the output they get:

![image](https://user-images.githubusercontent.com/198638/144534429-299879dc-284a-41d6-8b74-b0cffdbe1e17.png)

Also the full debug log from syslog.

```
Dec  3 02:02:47 hpt6 2021-12-03 02:02:47,735 INFO  [hpt] Decoded diagnostics report: {'diagnostics_passed': False, 'errors': ['ECC', 'BN', 'LOR'
, 'OK', 'PK', 'PF'], 'serial_number': '00000000b9c4a6f2', 'ECC': 'gateway_mfr test finished with error, {"result": "fail", "tests": [{"output": 
"ok", "result": "pass", "test": "serial"}, {"output": "unlocked", "result": "fail", "test": "zone_locked(data)"}, {"output": "unlocked", "result
": "fail", "test": "zone_locked(config)"}, {"output": "invalid slot: 0", "result": "fail", "test": "slot_config(0..=15, ecc)"}, {"output": "inva
lid slot: 3", "result": "fail", "test": "key_config(0..=15, ecc)"}, {"output": "ecc608 error\\n\\nCaused by:\\n    ecc error ExecError", "result
": "fail", "test": "miner_key(0)"}, {"output": "ecc608 error\\n\\nCaused by:\\n    ecc error ExecError", "result": "fail", "test": "sign(0)"}, {
"output": "ecc608 error\\n\\nCaused by:\\n    ecc error ExecError", "result": "fail", "test": "ecdh(0)"}]}', 'E0': 'F0:4C:D5:5A:52:73', 'eth_mac
_address': 'F0:4C:D5:5A:52:73', 'W0': '34:C3:D2:37:E0:72', 'wifi_mac_address': '34:C3:D2:37:E0:72', 'BN': 'Env var BALENA_DEVICE_NAME_AT_INIT no
t set', 'BALENA_DEVICE_NAME_AT_INIT': 'Env var BALENA_DEVICE_NAME_AT_INIT not set', 'ID': '61e1191213e5ef678b71ec612c4128a0', 'BALENA_DEVICE_UUI
[...]

PR #265, opened 2021-12-03
provision_key was never getting called 

kashifpk:
Provision_key was never getting called because if was in main() function while gunicorn via docker was calling the get_app function.

**Issue**
No new miners passing manufacturing.

- Link: https://github.com/NebraLtd/hm-diag/issues/264

### DIFF
--- a/hw_diag/app.py
+++ b/hw_diag/app.py
@@ -45,4 +45,3 @@ def get_app(name):
     app.register_blueprint(DIAGNOSTICS)
 
     return app
-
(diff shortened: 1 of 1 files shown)


vpetersson (approved):

hw_diag/app.py
```diff
@@ -26,6 +26,7 @@ def perform_key_provisioning():
 
 
 def get_app(name):
+    perform_key_provisioning()
```
vpetersson, line 29: Try

```
if os.getenv('BALENA_DEVICE_TYPE', False):
    perform_key_provisioning()
```

Issue #239, opened 2021-11-21
gateway_mfr exited with a non-zero status not handled well


shawaj:
This is another error message which is not handled well and stops the whole page and JSON from returning:

![Screenshot_20211121-020316_Chrome](https://user-images.githubusercontent.com/3359418/142746600-fa6a510a-efdb-4ebc-bbae-eb8a46e5decb.jpg)

We should audit all possible error messages to stop this from occurring in any circumstances

Related to:
- https://github.com/NebraLtd/hm-diag/issues/243 (with balena links)
- https://github.com/NebraLtd/hm-diag/issues/257

Need to determine if this is a hardware issue or software malfunction.

**Acceptance Criteria**
- Improve on-screen error message
- Improve logging
- Provide some sort of root cause analysis of why this happens

shawaj:
Weirdly I seem to get this on a device where gateway_mfr has succeeded and tests show all good, but the miner cache.dets is corrupted and repairing

shawaj:
Closed by https://github.com/NebraLtd/helium-miner-software/pull/311

PR #270, opened 2021-12-10
Handle gateway mfr exit code exception

SebastianMaj:
**Issue**
- Link: https://github.com/NebraLtd/hm-diag/issues/239
- Summary: Gateway MFR Error causing diagnostics page to not open or create a boot loop on the container.

**How**
Traced back the exception and added handling to the source of the issue. This lets the diagnostics continue gracefully and not raise an exception but instead handle it and display the diagnostics page but with errors displayed on diag view. 

**Screenshots**
Screen Shot 2021-12-10 at 2 57 32 AM

**References**
The screenshot above is what the dashboard looks like with the error handling when ECC fails from the error "Gateway mfr exited with non-zero status"

How to recreate this error locally:
In Balena dashboard set i2c_arm to false and as a result this will trigger the ECC to fail as the bus is inactive. Previously this triggered a boot loop on the diagnostics container, now it handles the error and displays it on the diagnostics page.

**Checklist**

- [x] Tests added
- [x] Documentation added

### DIFF
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -28,7 +28,13 @@
         <div class="text-center pb-4 mb-4 border-bottom border-light">
             <h1>Diagnostics Information</h1>
             <br />
-            <h3>{{ diagnostics.AN }}</h3>
+            <h3>
+              {% if diagnostics.AN == "ECC failure" %}
+                Animal Name Unavailable
+              {% else %}
+                {{ diagnostics.AN }}
+              {% endif %}
+            </h3>
             {% if diagnostics.PF %}
                     <h3 class="text-success">All Ok</h3>
             {% else %}
@@ -41,8 +47,13 @@
             <div class="card mb-0 h-100">
               <table class="table dt-responsive nowrap m-2 w-auto">
                 <tr>
-                  <td class="border-0">Helium Address</td>
-                  <td class="border-0 text-right"><a href="https://explorer.helium.com/hotspots/{{ diagnostics.PK }}" target="_blank">View on Explorer</a></td>
+                  {% if diagnositcs.PK == "ECC failure" %}
+                    <td class="border-0">Helium Address</td>
+                    <td class="border-0 text-right"><a href="https://explorer.helium.com/hotspots/{{ diagnostics.PK }}" target="_blank">View on Explorer</a></td>
+                  {% else %}
+                    <td class="border-0">Helium Address</td>
+                    <td class="border-0 text-right">Animal Name Unavailable</td>
+                  {% endif %}
                 </tr>
                 <tr>
                   <td>Sync Percentage</td>
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -29,7 +29,7 @@
             <h1>Diagnostics Information</h1>
             <br />
             <h3>
-              {% if diagnostics.AN == "ECC failure" %}
+              {% if not diagnostics.AN %}
                 Animal Name Unavailable
               {% else %}
                 {{ diagnostics.AN }}
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -47,7 +47,7 @@
             <div class="card mb-0 h-100">
               <table class="table dt-responsive nowrap m-2 w-auto">
                 <tr>
-                  {% if diagnositcs.PK == "ECC failure" %}
+                  {% if diagnositcs.PK != "ECC failure" %}
                     <td class="border-0">Helium Address</td>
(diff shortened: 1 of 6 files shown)


hw_diag/utilities/hardware.py
```diff
@@ -250,12 +250,19 @@ def lora_module_test():
 
 
 def get_public_keys_and_ignore_errors():
-    public_keys = get_public_keys_rust()
-    if not public_keys:
-        error_msg = "ECC failure"
+    try:
+        public_keys = get_public_keys_rust()
+        if not public_keys:
+            error_msg = "ECC failure"
```
marvinmarnold, line 256: nit: might as well define this once at beginning of method so you can reuse it in the `except` clause too

marvinmarnold (changes requested):
I personally find it a little awkward to list `ECC failure` for the animal name if there is a failure. I think its cleaner to omit it entirely. Alternatively `Animal name unavailable` would be more clear.

hw_diag/utilities/hardware.py
SebastianMaj: Done! 😊

SebastianMaj: @marvinmarnold Just going to test this to make sure the page looks good when ECC failure occurs

shawaj:
Further to @marvinmarnold comment it'll also mean the helium address link doesn't work I think.

So we should show the same error there I guess

SebastianMaj:
> Further to @marvinmarnold comment it'll also mean the helium address link doesn't work I think.
> 
> So we should show the same error there I guess

Indeed it should, it gives us the interesting URL of "https://explorer.helium.com/hotspots/ECC%20failure" 😂

SebastianMaj:
image

SebastianMaj:
image

hw_diag/templates/diagnostics_page.html
```diff
@@ -41,8 +47,13 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
             <div class="card mb-0 h-100">
               <table class="table dt-responsive nowrap m-2 w-auto">
                 <tr>
-                  <td class="border-0">Helium Address</td>
-                  <td class="border-0 text-right"><a href="https://explorer.helium.com/hotspots/{{ diagnostics.PK }}" target="_blank">View on Explorer</a></td>
+                  {% if diagnositcs.PK == "ECC failure" %}
+                    <td class="border-0">Helium Address</td>
+                    <td class="border-0 text-right"><a href="https://explorer.helium.com/hotspots/{{ diagnostics.PK }}" target="_blank">View on Explorer</a></td>
```
shawaj, line 52: Think this is the wrong way around

requirements.txt
```diff
@@ -7,5 +7,5 @@ gunicorn==20.1.0
 requests==2.26.0
 retry==0.9.2
 sentry-sdk==1.1.0
-hm-pyhelper==0.11.13
 dbus-python==1.2.16
+hm_pyhelper==0.12.0
```
shawaj, line 11: Think this should be hm-pyhelper and there's a v0.12.1 now too FYI

SebastianMaj: Oh, this is interesting, it worked with hm_pyhelper as it didn't fail on me. I'll change it to hm-pyhelper though as that's the correct name for the package.

hw_diag/templates/diagnostics_page.html
SebastianMaj: Thank you for bringing this up, I changed it on the device via SSH but forgot to change it in the code after testing

requirements.txt
shawaj: @SebastianMaj actually @kashifpk said it would work fine with underscore instead as it uses them somewhat interchangeably but just thought the "proper" name is better and it's what we use in all other repos AFAIK so just for consistency

shawaj (approved):

hw_diag/templates/diagnostics_page.html
```diff
@@ -28,7 +28,13 @@ <h3>{{ diagnostics.error }}</h3>
         <div class="text-center pb-4 mb-4 border-bottom border-light">
             <h1>Diagnostics Information</h1>
             <br />
-            <h3>{{ diagnostics.AN }}</h3>
+            <h3>
+              {% if diagnostics.AN == "ECC failure" %}
```
marvinmarnold, line 32: Is there existing logic that assumes `AN` will be `ECC failure`? If not, can we use `None` instead? That would be more consistent with how we are handling Helium API failures:

SebastianMaj: It seems like we are setting it to ECC failure [here](https://github.com/NebraLtd/hm-diag/blob/c22a42915cb4f029dae41f31076d9867f84894dd/hw_diag/utilities/hardware.py#L255)

I can change it to be ```None``` here I reviewed the rest of the references to public keys in hm_diag and it seems like this is the only spot we provide a different result than the original. Then in hm_pyhelper it seems like it solely relies on gateway_mfr to return the public keys result

marvinmarnold (approved):

Issue #273, opened 2021-12-19
Healthcheck fails - wget: not found


iankaufmann:
Firmware Version | 2021.12.14.0

I noticed that the diagnostics container was crashing often, and always in a (Starting) state.

Digging into `balena-engine inspect` I saw this:

```
"Health": {
    "Status": "starting",
    "FailingStreak": 4,
    "Log": [
        {
            "Start": "2021-12-19T01:48:09.391672934Z",
            "End": "2021-12-19T01:48:10.755888085Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        },
        {
            "Start": "2021-12-19T01:50:17.94280988Z",
            "End": "2021-12-19T01:50:19.17836156Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        },
        {
            "Start": "2021-12-19T01:52:19.216071717Z",
            "End": "2021-12-19T01:52:20.496799759Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        },
        {
            "Start": "2021-12-19T01:54:20.536884687Z",
            "End": "2021-12-19T01:54:21.996848867Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        },
        {
            "Start": "2021-12-19T01:56:22.043875034Z",
            "End": "2021-12-19T01:56:23.205485664Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        }
    ]
}
```

I'm not sure if this is causing other issues, but it appeared to be causing the miner to go unresponsive at least once an hour and until I enabled SSH to start poking around, I was having to power cycle it to get it to come back (for a short time).

After `exec` into the container and installing `wget` the container stopped crashing and reports (Healthy) state:

```
"Health": {
    "Status": "healthy",
    "FailingStreak": 0,
    "Log": [
        {
            "Start": "2021-12-19T01:58:23.246580687Z",
            "End": "2021-12-19T01:58:24.725130115Z",
            "ExitCode": 1,
            "Output": "/bin/sh: 1: wget: not found\n"
        },
        {
            "Start": "2021-12-19T02:00:24.773121117Z",
            "End": "2021-12-19T02:00:27.073970549Z",
            "ExitCode": 0,
[...]

PR #274, opened 2021-12-22
Adds wget for healthcheck

vpetersson:
**Issue**
Closes https://github.com/NebraLtd/hm-diag/issues/273

**How**
In theory, the [healthcheck](https://github.com/NebraLtd/hm-diag/blob/master/Dockerfile#L35-L40) shouldn't really do anything on Balena, since it hasn't been fully implemented. However, as part of the logs in [273](https://github.com/NebraLtd/hm-diag/issues/273) it does indeed seems like they now have been. 

It's still clear if these healthchecks actually do anything, but at least it shouldn't cause any harm to resolve this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
 
 RUN \
     install_packages \
+        wget \
         i2c-tools \
         libdbus-1-3
 


SebastianMaj (approved):
Looks good!

ilyastrodubtsev (approved):

shawaj:
@vpetersson iirc the balena raspberry pi base images already have wget installed...

vpetersson:
That's what I thought too but it seems to have changed. See the referenced issue. Seems like it isn't.

shawaj:
@vpetersson did you check though on live miners? Given the customer has modified their miner I'm not convinced it's an issue on our side.

PR #291, opened 2022-01-25
feat: Added short Git hash report (hm-diag#205)

MuratUrsavas:
Include short git-hash report.

- Link: hm-diag#205 
- Summary:

We needed to include short git hash of main repo (helium-miner-software) for release tracking purposes.

**How**
- The short git hash of main repo is fetched via an environment variable supplied by the main docker container composer (helium-miner-software)
- The hash is propagated via multiple routes (/json, /version, /)
- Short code for Short Git Hash (SGH) has been introduced.

**Checklist**

- [x] Tests added
- [x] Documentation added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# VSCode editor related stuff
+.vscode/
--- a/hw_diag/diagnostics/env_var_diagnostics.py
+++ b/hw_diag/diagnostics/env_var_diagnostics.py
@@ -19,6 +19,9 @@ ENV_VARS_MAPPING = [{
 }, {
     'ENV_VAR': 'VARIANT',
     'DIAGNOSTIC_KEY': 'VA'
+}, {
+    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
+    'DIAGNOSTIC_KEY': 'firmware_short_hash'
 }]
 
 
--- a/hw_diag/tests/diagnostics/test_env_var_diagnostics.py
+++ b/hw_diag/tests/diagnostics/test_env_var_diagnostics.py
@@ -56,5 +56,10 @@ class TestEnvVarDiagnostics(unittest.TestCase):
             'FIRMWARE_VERSION': 'foo',
             'FW': 'foo',
             'VARIANT': 'foo',
-            'VA': 'foo'
+            'VA': 'foo',
+            # We're moving towards longer lowercase key naming and will
+            # deprecate old ones in near future. Just keeping this entry in the
+            # list for the sake of style compatibility.
+            'FIRMWARE_SHORT_HASH': 'foo',
+            'firmware_short_hash': 'foo'
         })
--- a/hw_diag/tests/test_views_diagnostics.py
+++ b/hw_diag/tests/test_views_diagnostics.py
@@ -71,7 +71,11 @@ class TestGetDiagnostics(unittest.TestCase):
 
     @patch.dict(
         os.environ,
-        {'FIRMWARE_VERSION': '1337.13.37', 'DIAGNOSTICS_VERSION': 'aabbffe'}
+        {
+            'FIRMWARE_VERSION': '1337.13.37',
+            'DIAGNOSTICS_VERSION': 'aabbffe',
+            'FIRMWARE_SHORT_HASH': '0011223'
+        }
     )
     def test_version_endpoint(self):
         # Check the version json output
@@ -81,3 +85,4 @@ class TestGetDiagnostics(unittest.TestCase):
         self.assertEqual(resp.status_code, 200)
         self.assertEqual(reply['firmware_version'], '1337.13.37')
         self.assertEqual(reply['diagnostics_version'], 'aabbffe')
+        self.assertEqual(reply['firmware_short_hash'], '0011223')
--- a/hw_diag/utilities/shell.py
+++ b/hw_diag/utilities/shell.py
@@ -13,9 +13,10 @@ def get_environment_var(diagnostics):
         'BALENA_APP_NAME',
         'FREQ',
         'FIRMWARE_VERSION',
-        'VARIANT'
+        'VARIANT',
+        'FIRMWARE_SHORT_HASH'
     ]
-    keys = ["BN", "ID", "BA", "FR", "FW", "VA"]
+    keys = ["BN", "ID", "BA", "FR", "FW", "VA", "firmware_short_hash"]
 
     for (var, key) in zip(env_var, keys):
         diagnostics[key] = os.getenv(var)
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -83,7 +83,7 @@
                 </tr>
                 <tr>
                   <td>Firmware Version</td>
-                  <td class="text-right">{{ diagnostics.FW }}</td>
+                  <td class="text-right">{{ diagnostics.FW }} ({{ diagnostics.firmware_short_hash }})</td>
                 </tr>
                 <tr>
                   <td>Frequency</td>
--- a/hw_diag/views/diagnostics.py
+++ b/hw_diag/views/diagnostics.py
@@ -107,6 +107,7 @@ def version_information():
     response = {
         'firmware_version': os.getenv('FIRMWARE_VERSION', 'unknown'),
         'diagnostics_version': os.getenv('DIAGNOSTICS_VERSION', 'unknown'),
+        'firmware_short_hash': os.getenv('FIRMWARE_SHORT_HASH', 'unknown'),
     }
 
     return response


hw_diag/diagnostics/env_var_diagnostics.py
```diff
@@ -19,6 +19,9 @@
 }, {
     'ENV_VAR': 'VARIANT',
     'DIAGNOSTIC_KEY': 'VA'
+}, {
+    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
+    'DIAGNOSTIC_KEY': 'FSH'
```
kashifpk, line 24: @MuratUrsavas - Why are we adding this here? We can just display the hash in the `/version` endpoint. Don't think we need that in `/initFile.txt` payload.

hw_diag/tests/diagnostics/test_env_var_diagnostics.py
```diff
@@ -56,5 +56,7 @@ def test_env_vars_success(self):
             'FIRMWARE_VERSION': 'foo',
             'FW': 'foo',
             'VARIANT': 'foo',
-            'VA': 'foo'
+            'VA': 'foo',
+            'FIRMWARE_SHORT_HASH': 'foo',
+            'FSH': 'foo'
```
kashifpk, line 61: No need for the shorter `FSH` name. The short keys are only 2-characters long and we are moving away from them gradually. So instead of `VA` we try to use `VARIANT`. These legacy 2-character keys will be removed once all code using them has been updated to use the more friendlier keys.

hw_diag/tests/test_views_diagnostics.py
```diff
@@ -71,7 +71,11 @@ def test_initFile_output(self):
 
     @patch.dict(
         os.environ,
-        {'FIRMWARE_VERSION': '1337.13.37', 'DIAGNOSTICS_VERSION': 'aabbffe'}
+        {
```
kashifpk, line 74: Same as above, we don't need this in `/initFile.txt` payload.

hw_diag/utilities/shell.py
```diff
@@ -13,9 +13,10 @@ def get_environment_var(diagnostics):
         'BALENA_APP_NAME',
         'FREQ',
         'FIRMWARE_VERSION',
-        'VARIANT'
+        'VARIANT',
+        'FIRMWARE_SHORT_HASH'
```
kashifpk, line 17: Not needed as mentioned above.

hw_diag/utilities/shell.py
```diff
@@ -13,9 +13,10 @@ def get_environment_var(diagnostics):
...
     ]
-    keys = ["BN", "ID", "BA", "FR", "FW", "VA"]
+    keys = ["BN", "ID", "BA", "FR", "FW", "VA", "FSH"]
```
kashifpk, line 19: Not needed as mentioned above.

kashifpk (changes requested):
Most of the changes requested are around us not needing the versions and hashes in `/initFile.txt` payload. Rest looks good.

hw_diag/diagnostics/env_var_diagnostics.py
MuratUrsavas: I thought it would be nice to report the hash where the FIRMWARE_VERSION is reported. I didn't know the direction and what was going to be deprecated in the near future.

MuratUrsavas: Thought again, and we have to have this duo anyway, because the mapping has to include FIRMWARE_SHORT_HASH (We're getting it from an environment variable) and in the current scenario we need to give it a short name. If we'd remove the short naming scheme, then this would go away too. I making it two chars as FH.

hw_diag/tests/test_views_diagnostics.py
MuratUrsavas: @Kashif please correct me if I'm wrong. This line is not about testing the initFile.txt, it is for `/version` and we must report amd test it under `/version`. How can we run `test_version_endpoint()` without setting the environment var?

hw_diag/utilities/shell.py
MuratUrsavas: Like above, I think we need to keep the short name until we remove the short naming scheme completely.

hw_diag/tests/test_views_diagnostics.py
kashif: kindly refrain from randomly adding random usernames to your issue... I believe this is a typo?

hw_diag/utilities/shell.py
kashifpk: @MuratUrsavas  - No new code we write relies on the short 2-character names and we don't have those for any new keys in endpoints either. See `/version` endpoint, it does not have any short names. Also the short names you see in `/initFile.txt` payload are only kept not to break existing code the uses them. So if you're adding a new short name, by definition no code is using it hence there is no reason for it to be present.

MuratUrsavas: OK, this one is easy. I can remove it. Let's keep the other one for the reason I've mentioned earlier. Cool?

MuratUrsavas: Although, I still prefer to keep all of them just for consistency and remove them later altogether.

hw_diag/diagnostics/env_var_diagnostics.py
kashifpk: Ok this gets used in the web ui too so it's fine to keep it until we start using a different method for displaying this info in the web interface.

hw_diag/tests/test_views_diagnostics.py
kashifpk: > e must report amd test it under `/version`. How can we run `test_version_endpoint()` without setting the environment var?

Yes my bad, this is fine for version endpoint. And without addressing "Kashif Rasool", yes it was a typo.

hw_diag/diagnostics/env_var_diagnostics.py
```diff
@@ -19,6 +19,9 @@
 }, {
     'ENV_VAR': 'VARIANT',
     'DIAGNOSTIC_KEY': 'VA'
+}, {
+    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
+    'DIAGNOSTIC_KEY': 'FH'
```
marvinmarnold, line 24: `'DIAGNOSTIC_KEY': 'firmware_short_hash'` would be better

marvinmarnold (changes requested):

hw_diag/diagnostics/env_var_diagnostics.py
MuratUrsavas: Done.

kashifpk (approved):

ilyastrodubtsev (approved):

hw_diag/tests/diagnostics/test_env_var_diagnostics.py
```diff
@@ -56,5 +56,7 @@ def test_env_vars_success(self):
             'FIRMWARE_VERSION': 'foo',
             'FW': 'foo',
             'VARIANT': 'foo',
-            'VA': 'foo'
+            'VA': 'foo',
+            'FIRMWARE_SHORT_HASH': 'foo',
+            'firmware_short_hash': 'foo'
```
vpetersson, line 61: Why do we have both lowercase and upper case here?

MuratUrsavas: Good question. I've kept the existing format before and of course the last change made it look pretty ugly (only case change). Should I remove the upper case from here?

MuratUrsavas: Before that it was only `FH` for backward compatibility.

MuratUrsavas: Added the comment as we talked in the meeting.

.gitignore
```diff
@@ -137,3 +137,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# VSCode editor related stuff
+.vscode/
```
marvinmarnold, line 142: I have the slightest preference for dev specific things to go in your local global gitignore. Not worth undoing, but think we should avoid intentionally adding to the gitignore of all repos.

hw_diag/diagnostics/env_var_diagnostics.py
```diff
@@ -19,6 +19,9 @@
 }, {
     'ENV_VAR': 'VARIANT',
     'DIAGNOSTIC_KEY': 'VA'
+}, {
```
marvinmarnold, line 22: I think part of the confusion was that this array structure is pretty confusing (I wrote it originally, whoops). You have to dig through the code to understand that the `ENV_VAR` is used as the friendly_key and that `DIAGNOSTIC_KEY` is the key. 

Not worth creating a ticket just to refactor, but the next time one of us touches this code, it might be worth refactoring to something like:

```
[
...
  {
    'ENV_VAR': 'FREQ',
    'DIAGNOSTIC_KEY': 'FR'
    'DIAGNOSTIC_FRIENDLY_KEY': 'frequency'
  }
]
```

What do you think?

hw_diag/diagnostics/env_var_diagnostics.py
```diff
@@ -19,6 +19,9 @@
...
+    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
```
marvinmarnold, line 23: I think part of the confusion was that this array structure is hard to understand (I wrote it originally, whoops). You have to dig through the code to discover that the `ENV_VAR` is used as the friendly_key and that `DIAGNOSTIC_KEY` is the key. 

Not worth creating a ticket just to refactor, but the next time one of us touches this code, it might be worth refactoring to something like:

```
[
...
  {
    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
    'DIAGNOSTIC_KEY': 'firmware_short_hash
    'DIAGNOSTIC_FRIENDLY_KEY': 'firmware_short_hash'
  }
]
```

@kashifpk taking it a step towards https://github.com/NebraLtd/hm-pyhelper/issues/104, it could be something like:

```
from hm_pyhelper.constants import DIAGNOSTICS_KEYS

[
...
  {
    'ENV_VAR': 'FIRMWARE_SHORT_HASH',
    'DIAGNOSTIC_KEY': DIAGNOSTICS_KEYS.firmware_short_hash,
    'DIAGNOSTIC_FRIENDLY_KEY': DIAGNOSTICS_KEYS.firmware_short_hash
  }
]
```

What do you think?

marvinmarnold (approved):

marvinmarnold:
Nothing blocking merge.

.gitignore
kashifpk: Learned something new and just in time @marvinmarnold :-) Was thinking whether to add my local docker-compose files (leaner versions with fewer services) for hm-dashboad to hm-dashboard's .gitignore and not and found your comment just in time.

FYI for anyone interested on *nix and mac you can add such patterns to `~/.config/git/ignore`

hw_diag/diagnostics/env_var_diagnostics.py
kashifpk: @marvinmarnold - yes, the only thing I'd change would be to have the constants in all upper case. so something like

```python
from hm_pyhelper.constants.diagnostics import FIRMWARE_SHORT_HASH_KEY
```

Already using similar logic in https://github.com/NebraLtd/Hotspot-Production-Tool/blob/master/app/src/hpt/constants.py

These would get moved to https://github.com/NebraLtd/hm-pyhelper/issues/104 once we tackle that issue.

.gitignore
MuratUrsavas: Learned another thing today. That's a `+1` for me :smiley:

Issue #283, opened 2022-01-13
Include AddGateway transaction in /initFile payload


marvinmarnold:
Pass this through to the dashboard's registration payload.

Ensure writability: mount | grep boot
/dev/mmcblk0p1 on /mnt/boot type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro)

No longer require nebra.json to be on device. HPT will send a request with the necessary payload.

**Acceptance Criteria**

- Add Nebra's manufacturing public key corresponding to ihub.json.asc.
- POST /v1/add_gateway_txn should:
  - Should accept ihub.json.asc as a param. Either as string contents, or file. 
  - If ihub.json.asc's signature is validated, continue, otherwise return error.   
  - If valid, generate `shipping_destination_add_gateway_txn` by invoking https://github.com/NebraLtd/hm-pyhelper/issues/91 and supplying a random wallet from the list.
   - Always return Diagnostics report as payload.
   - Return 404 if there is an error.
   - `shipping_destination_add_gateway_txn` must be a JSON that includes the transaction and wallet ID used, presumably it [should look like](https://docs.helium.com/mine-hnt/full-hotspots/become-a-maker/hotspot-integration-testing#generate-an-add-hotspot-transaction):

```
{
  "address": "11TL62V8NYvSTXmV5CZCjaucskvNR1Fdar1Pg4Hzmzk5tk2JBac",
  "fee": 65000,
  ~"mode": "full",~ (ignore, not relevant)
  "owner": "14GWyFj9FjLHzoN3aX7Tq7PL6fEg4dfWPY8CrK8b9S5ZrcKDz6S",
  "payer": "138LbePH4r7hWPuTnK6HXVJ8ATM2QU71iVHzLTup1UbnPDvbxmr",
  "staking fee": 4000000,
[...]

PR #292, opened 2022-01-25
Support nebra.json and add_gateway_txn

SebastianMaj:
**Issue**

- Link: #283 
- Summary: Create diagnostics and API endpoints for nebra.json and add_gateway_txn.

**How**
- Created NebraJsonDiagnostics to copy values from nebra.json to a DiagnosticsReport.
- Created GenAddGatewayTxn and AckAddGatewayTxn to generated an add_gateway_txn and to remove destination wallets once complete.
- Updated tests that expect friendly_keys to be in errors array following pyhelper update.

**References**
[Generation of Nebra.json file](https://github.com/NebraLtd/hotspot-production-images/issues/26)
[hm-pyhelper JSON RPC function for generation of gateway transaction](https://github.com/NebraLtd/hm-pyhelper/issues/91)

**Checklist**

- [x] Tests added
- [x] Cleaned up commit history (rebase!)
- [ ] Documentation added
- [x] Thought about variable and method names

PR #298, opened 2022-02-14
feat: Implement add-gateway-txn endpoint

posterzh:
**Issue**

- Link: https://github.com/NebraLtd/hm-diag/issues/283
- Summary: 
Create API endpoints for add_gateway_txn.

**How**
- Export gpg key at **HPT** side:
`gpg --armor --output key-file-pub.gpg --export 0xBFBFA70B`
- Put exported [key](https://github.com/NebraLtd/hm-diag/blob/f2eb857b503280afcc6d1af99880280eb3284503/keys/key-file-pub.gpg) in  **hm-diag**. Update the [key path](https://github.com/NebraLtd/hm-diag/pull/298/files#diff-f83c178edbd57f2cd4ac05b89e9a1ead14096c230ab4114899ad5e65175316ccR6) constant accordingly(`KEY_PATH = "./keys/key-file-pub.gpg"`).
- Create `/v1/add-gateway-txn` POST API endpoint in  **hm-diag**.
  - Decrypt and validate the file payload using [python-gnupg](https://pypi.org/project/python-gnupg/)
  - Call miner RPC api by https://github.com/NebraLtd/hm-pyhelper/issues/91
- Encrypt the client JSON file at **HPT** side. The following command will produce `shipping_destination.json.asc` file.
`gpg -u 0xBFBFA70B --clearsign shipping_destination.json`
- Call `/v1/add-gateway-txn` REST API endpoint with POST method, file field=shipping_destination.json.asc
- `/v1/add-gateway-txn` endpoint in `hm-diag` will create `add-gateway-txn` transaction and submit to the Helium blockchain. And it will return the response with a descriptive message and http status code(200/406)

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


hw_diag/utilities/security.py
```diff
@@ -0,0 +1,37 @@
...
+
+    LOGGER.debug(f'Importing {KEY_PATH} failed.')
+    return None
+
+
+# Decrypt file using the gnupg
+# Return values:
+#   Decrypted data if successful
+#   None if decryption is unsuccessful
+def decrypt_file(file):
```
marvinmarnold, line 28: Should be verification, not decryption: https://docs.red-dove.com/python-gnupg/#verification

marvinmarnold (changes requested):

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +118,43 @@ def version_information():
...
+            "address": "11TL62V8NYvSTXmV5CZCjaucskvNR1Fdar1Pg4Hzmzk5tk2JBac",
+            "fee": 65000,
+            "owner": "14GWyFj9FjLHzoN3aX7Tq7PL6fEg4dfWPY8CrK8b9S5ZrcKDz6S",
+            "payer": "138LbePH4r7hWPuTnK6HXVJ8ATM2QU71iVHzLTup1UbnPDvbxmr",
+            "staking fee": 4000000,
+            "txn": "CrkBCiEBrlImpYLbJ0z0hw5b4g9isRyPrgbXs9X+RrJ4pJJc9MkS..."
+        }
+    }
+    """
+    file = request.files['file']
```
marvinmarnold, line 142: Implementation should look more like this: https://github.com/NebraLtd/hm-diag/pull/292/files#diff-04b1aca2fd0e3d6dca6921bc407e7e1cf9b6c93e823d1d31591fcfd6b398304aR114

hw_diag/tests/test_views_diagnostics.py
```diff
@@ -86,3 +90,65 @@ def test_version_endpoint(self):
         self.assertEqual(reply['firmware_version'], '1337.13.37')
         self.assertEqual(reply['diagnostics_version'], 'aabbffe')
         self.assertEqual(reply['firmware_short_hash'], '0011223')
+
+    CLIENT_JSON_ENCRYPTED = b"""
```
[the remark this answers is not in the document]
posterzh: clear name

hw_diag/utilities/security.py
posterzh: Make sense.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,42 @@
...
+from hm_pyhelper.constants.shipping import DESTINATION_WALLETS_KEY
+from hm_pyhelper.miner_json_rpc.exceptions import MinerMalformedAddGatewayTxn
+
+from hw_diag.utilities.miner import create_add_gateway_txn
+from hw_diag.utilities.security import get_verified_data
+
+KEY = 'ADD_GATEWAY_TXN'
+FRIENDLY_NAME = "add-gateway-transaction"
+
+NO_DESTINATION_WALLETS_MSG = "Destination wallets not found in payload JSON."
```
marvinmarnold, line 14: If I understand @kashifpk, `NO_DESTINATION_WALLETS_MSG` and `VERIFICATION_FAILED_MSG` should be class variable since they won't be accessed externally.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,42 @@
...
+
+from hm_pyhelper.diagnostics import Diagnostic
+from hm_pyhelper.constants.shipping import DESTINATION_WALLETS_KEY
+from hm_pyhelper.miner_json_rpc.exceptions import MinerMalformedAddGatewayTxn
+
+from hw_diag.utilities.miner import create_add_gateway_txn
+from hw_diag.utilities.security import get_verified_data
+
+KEY = 'ADD_GATEWAY_TXN'
+FRIENDLY_NAME = "add-gateway-transaction"
```
marvinmarnold, line 12: Use this constant from pyhelper: https://github.com/NebraLtd/hm-pyhelper/blob/master/hm_pyhelper/constants/shipping.py#L2

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,42 @@
...
+import secrets
+
+from hm_pyhelper.diagnostics import Diagnostic
+from hm_pyhelper.constants.shipping import DESTINATION_WALLETS_KEY
+from hm_pyhelper.miner_json_rpc.exceptions import MinerMalformedAddGatewayTxn
+
+from hw_diag.utilities.miner import create_add_gateway_txn
+from hw_diag.utilities.security import get_verified_data
+
+KEY = 'ADD_GATEWAY_TXN'
```
marvinmarnold, line 11: KEY and FRIENDLY_NAME should be the same thing. They are only different to support legacy code that used an ugly format.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,42 @@
...
+FRIENDLY_NAME = "add-gateway-transaction"
+
+NO_DESTINATION_WALLETS_MSG = "Destination wallets not found in payload JSON."
+VERIFICATION_FAILED_MSG = "Verifying the payload signature failed."
+
+
+class AddGatewayTxnDiagnostic(Diagnostic):
```
marvinmarnold, line 18: Add type hints everywhere

.github/workflows/python-tests.yml
```diff
@@ -14,7 +14,9 @@ jobs:
       with:
         python-version: 3.8
     - name: Install apt dependencies
-      run: sudo apt-get install libdbus-1-dev libdbus-glib-1-dev python3-dev
+      run: sudo apt-get install libdbus-1-dev libdbus-glib-1-dev python3-dev gpg
+    - name: Import gpg key
+      run: gpg --import keys/key-file-pub.gpg
```
marvinmarnold, line 19: Tests should use a special key. We don't want to have to update all tests if we rotate the key for some business reason.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, unverified_shipping_destination):
+        super(AddGatewayTxnDiagnostic, self).__init__(KEY, FRIENDLY_NAME)
+        self.unverified_shipping_destination = unverified_shipping_destination
```
marvinmarnold, line 21: `unverified_shipping_destination_file`

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +113,40 @@ def version_information():
...
```
marvinmarnold, line 137: Are we handling errors on lines 137 or 138? Should return a clear error if the request is invalid.

hw_diag/utilities/security.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def get_verified_data(data):
+    """ Verify the signature on the contents of the string 'data'
+        Return values:
+            Verified data - if succeed
+            None - if verification failed
+    """
+    verified = gpg.verify(data)
+    if verified.valid:
```
marvinmarnold, line 13: I was wrong about this. `trust_text` is for something else. The old way you had it was fine. We do have to use `decrypt` in either case. That is the part I think is most confusing.

Could you add a comment to explain that this decryption is just separating the payload from the rest of the signature and not doing any real decryption because the message is sent in cleartext?

marvinmarnold (changes requested):

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
marvinmarnold: @ilyastrodubtsev can you update [hm-dashboard](https://github.com/NebraLtd/hm-dashboard/blob/081077c2a75b9295265ae85101a3c17362db8e0c/app/device/views.py#L614) to reference the same constant: https://github.com/NebraLtd/hm-pyhelper/blob/27c5c2112475bc1a658bd5d98ecae8d2b862454a/hm_pyhelper/constants/shipping.py#L2

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
kashifpk: Yes @marvinmarnold - That's just to avoid namespace collisions if the same name gets used elsewhere and we import both of these in a single module. @posterzh placing them inside the class  ensures that they can't be  imported to global namespace with `from add_gateway_txn_diiagnostic import *`. Although using `from module import *` isn't something we do or should do.

This is being just a bit more careful :-)

.github/workflows/python-tests.yml
kashifpk: Yes @posterzh - as a general rule we try to avoid using actual input files for tests. In this case you can use gpg to [create a new key pair](https://github.com/NebraLtd/Hotspot-Production-Tool/tree/kashifpk/137-shipping-destinations#create-key-pair) during test setup. Or just generate one and store it with test cases that will only be used for tests.

This is going to be a bit tricky because python-gpg library used the actual gpg binary which assumes the default home folder (i.e, --homedir ~/.gnupg). Ideally you'd want to set the gpg homedir before running tests so it uses the keys you have either generated as part of test setup or already stored with tests. Since the library already supports this by creating the gpg object like:

```python
gpg = gnupg.GPG(gnupghome='/path/to/home/directory')
```

you can use this for testing assuming you update your code [here](https://github.com/NebraLtd/hm-diag/blob/126d57ed823dd3078269d40caeecdfb4e8e101ca/hw_diag/utilities/security.py#L3) to also accept a homedir path which if not present should default to `~/.gnupg`. Also another note, the gpg object you create should probably be created inside of the function.

hw_diag/utilities/security.py
```diff
@@ -0,0 +1,17 @@
+import gnupg
+
+gpg = gnupg.GPG()
```
kashifpk, line 3: Perhaps this should be moved inside the function and also should accept a homedir path to make testing easier?

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
posterzh: Make sense.

hw_diag/views/diagnostics.py
kashifpk, line 137: @posterzh - This looks like HTTP `multipart/form-data` style form POST. Where you are expecting the file to be present inside the file key. I was thinking more on the lines of reading the full raw POST data as the file content. On Flask side this means using `request.get_data()` or `request.data` instead of using `request.files` or `request.form`. See https://tedboy.github.io/flask/generated/generated/flask.Request.get_data.html#flask.Request.get_data

posterzh: Ok, it seems that `request.get_data()` can get the content of the raw file correctly without breaking the encoding. @kashifpk  Could you give show me how the client code that calls this end point should look like?

.github/workflows/python-tests.yml
posterzh: I agree to use gpg test key for test cases. I think it can be done without using custom homedir.

hw_diag/utilities/security.py
posterzh: Sure

hw_diag/views/diagnostics.py
kashifpk: Sure:

```python
import requests
my_data = 'your full payload with signature and actual json'.decode('utf8')  # converting to bytes from str
requests.post(url, data=my_data)
```

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,40 @@
...
+from hw_diag.utilities.miner import create_add_gateway_txn
+from hw_diag.utilities.security import get_verified_json
+
+
+class AddGatewayTxnDiagnostic(Diagnostic):
+    # Error messages
+    NO_DESTINATION_WALLETS_MSG = "Destination wallets not found in payload JSON."
+    VERIFICATION_FAILED_MSG = "Verifying the payload signature failed."
+
+    def __init__(self, shipping_destination_with_signature):
```
marvinmarnold, line 16: missing type hint

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
```diff
@@ -0,0 +1,40 @@
...
+        if DESTINATION_WALLETS_KEY not in shipping_destination_json:
+            diagnostics_report.record_failure(self.NO_DESTINATION_WALLETS_MSG, self)
+            return
+
+        destination_wallet = secrets.choice(shipping_destination_json[DESTINATION_WALLETS_KEY])
+
+        try:
+            add_gateway_txn = create_add_gateway_txn(destination_wallet)
+            diagnostics_report.record_result(add_gateway_txn, self)
+        except MinerMalformedAddGatewayTxn as e:
```
marvinmarnold, line 37: If we treat both errors the same, I don't think there is a benefit of explicitly catching `MinerMalformedAddGatewayTxn` is there?

hw_diag/diagnostics/lte_diagnostic.py
```diff
@@ -54,45 +56,43 @@ def perform_test(self, diagnostics_report):
...
+        manager = dbus.Interface(proxy, self.DBUS_OBJECTMANAGER)
+        modems = manager.GetManagedObjects()
+
+        for modem in modems:
+            self.append_lte_device_from_modem(lte_devices, modem, system_bus)
+
+        LOGGER.info(f"Found the following LTE Devices: {lte_devices}")
+        return lte_devices
+
+    def append_lte_device_from_modem(self, lte_devices: list, modem, system_bus) -> list:
```
marvinmarnold, line 78: Missing some type hints

hw_diag/diagnostics/mac_diagnostics.py
```diff
@@ -1,27 +1,14 @@
...
-  {
-    'key': 'W0',
-    'friendly_key': 'wifi_mac_address',
-    'mac_filepath': '/sys/class/net/wlan0/address'
-  }
-]
-
-
 class MacDiagnostic(Diagnostic):
     def __init__(self, key, friendly_key, mac_filepath):
```
marvinmarnold, line 7: Missing type hints

hw_diag/tests/data/test-key-pub.gpg
```diff
@@ -0,0 +1,23 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
marvinmarnold, line 1: If you are 1000000% positive this is a test key not used anywhere else, you might as well add the private key too, in case we every want to use it for any tests in the future.

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +113,44 @@ def version_information():
...
+            "staking fee": 4000000,
+            "txn": "CrkBCiEBrlImpYLbJ0z0hw5b4g9isRyPrgbXs9X+RrJ4pJJc9MkS..."
+        }
+    }
+    """
+
+    shipping_destination_with_signature = request.get_data()
+    if not shipping_destination_with_signature:
+        return {
+            'Error': 'Can not find payload.'
```
marvinmarnold, line 141: This should also return a DiagnosticsReport or client code will need to be more complex.

marvinmarnold (changes requested):
The only blocking comment is that the API endpoint should ALWAYS return a DiagnosticsReport. Thanks for refactoring code you didn't have to, that's helpful.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
posterzh: I agree.

hw_diag/tests/data/test-key-pub.gpg
posterzh: I am sure a test key won't be used elsewhere. The test key can't compromise the security and it has no relation with the official key. So I am going to add the private key that correspond to the test public key.

hw_diag/utilities/security.py
marvinmarnold: I was wrong about this. There is no way to extract the relevant payload without using decrypt.

posterzh: Now, updated to use `decrypt()` method but with a semantic name.

hw_diag/tests/test_views_diagnostics.py
```diff
@@ -86,3 +87,27 @@ def test_version_endpoint(self):
         self.assertEqual(reply['firmware_version'], '1337.13.37')
         self.assertEqual(reply['diagnostics_version'], 'aabbffe')
         self.assertEqual(reply['firmware_short_hash'], '0011223')
+
+    @patch('hw_diag.diagnostics.add_gateway_txn_diagnostic.AddGatewayTxnDiagnostic.perform_test')
+    def test_add_gateway_txn_success(self, mock):
```
marvinmarnold, line 92: We haven't really talked about this as a team, but for API endpoints, I think we should try to implement integration tests when possible. In this case, it would mean not mocking the diagnostic. You don't need to change this, but a thought for the future.

hw_diag/tests/test_views_diagnostics.py
```diff
@@ -86,3 +87,27 @@ def test_version_endpoint(self):
...
+
+    @patch('hw_diag.diagnostics.add_gateway_txn_diagnostic.AddGatewayTxnDiagnostic.perform_test')
+    def test_add_gateway_txn_failure_no_file_payload(self, mock):
+        endpoint = '/v1/add-gateway-txn'
+        resp = self.client.post(endpoint, data=None)
+
+        self.assertEqual(resp.status_code, 406)
+        self.assertDictEqual(resp.json, {
+            'diagnostics_passed': False,
+            'errors': ['error_message', 'error_message'],
```
marvinmarnold, line 111: I find this kind of confusing. The keys normally indicate some kind of scope about the information, but `error_message` is confusing out of context. I think the key should still be `DESTINATION_ADD_GATEWAY_TXN_KEY`.

keys/key-file-pub.gpg
```diff
@@ -0,0 +1,23 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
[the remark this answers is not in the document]
marvinmarnold: @vpetersson this is the only major blocker to merging now FYI.

hw_diag/utilities/security.py
posterzh: After some thoughts, the unit tests are made to use a test key without accepting a custom homedir.
However, accepting a homedir will be useful in case of when I couldn't make testing work otherwise.
Thank you.

hw_diag/tests/test_views_diagnostics.py
posterzh: Make sense. Ok, I am going to use (key, err_msg) as the parameter.

hw_diag/tests/test_views_diagnostics.py
posterzh: Good advice :+1: ! As a best practice, I will add both of the unit test(for ex, mocking the diagnostic) and the integration test(for ex, not mocking the diagnostic or mocking the lower level such as miner API) in the future.

hw_diag/views/diagnostics.py
posterzh: Fair. Updated now.

hw_diag/diagnostics/add_gateway_txn_diagnostic.py
posterzh: Now, all the methods of the all diagnostics are type hinted.

hw_diag/diagnostics/lte_diagnostic.py
posterzh: Now, all the methods of the all diagnostics are type hinted.

hw_diag/diagnostics/mac_diagnostics.py
posterzh: Now, all the methods of the all diagnostics are type hinted.

hw_diag/utilities/security.py
kashifpk: > without using

[get_payload_from_clearsigned_message](https://github.com/NebraLtd/hm-pyhelper/blob/master/hm_pyhelper/util/pgp.py#L4) would beg to differ :-) But using decrypt here is the better (and more reliable) choice since GPG and python-gpg are already dependencies for hm-diag.

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +114,46 @@ def version_information():
...
+@DIAGNOSTICS.route('/v1/add-gateway-txn', methods=['POST'])
+def add_gateway_txn():
+    """
+    Generates an add_gateway_txn if a destination name and wallets are defined and valid.
+    Diagnostics report will be in the format below if successful.
+
+    https://docs.helium.com/mine-hnt/full-hotspots/become-a-maker/hotspot-integration-testing/#generate-an-add-hotspot-transaction
+    {
+        errors: [],
+        DESTINATION_ADD_GATEWAY_TXN_KEY: {
```
kashifpk, line 128: @posterzh - The actual payload would contain the actual key name `shipping_destination_add_gateway_txn` here right? because `DESTINATION_ADD_GATEWAY_TXN_KEY` itself is a constant from `hm_pyhelper.constants.shipping`. Can you confirm this?

posterzh: Correct. `DESTINATION_ADD_GATEWAY_TXN_KEY` comes from `hm_pyhelper.constants.shipping` and its value is `shipping_destination_add_gateway_txn`.

Dockerfile
```diff
@@ -31,7 +31,15 @@ RUN \
...
-        libdbus-1-3
+        libdbus-1-3 \
+        gpg
+
+# Nebra uses /opt by convention
+WORKDIR /opt/
+
+# Import gpg key
+COPY keys/manufacturing-key.gpg ./
+RUN gpg --import manufacturing-key.gpg
```
kashifpk, line 42: @posterzh - once imported should we remove the key from /opt? Although keeping it there also does not affect anything.

hw_diag/diagnostics/bt_diagnostic.py
```diff
@@ -1,33 +1,34 @@
...
+LOGGER = get_logger(__name__)
 
-# BlueZ
-DBUS_BLUEZ_SERVICE_NAME = 'org.bluez'
-DBUS_ADAPTER_IFACE = 'org.bluez.Adapter1'
 
-NO_BT_DEVICES_MSG = "Bluez is working but, no Bluetooth devices detected."
+class BtDiagnostic(Diagnostic):
+    # Diagnostics keys
+    KEY = 'BT'
```
kashifpk, line 12: thumbs up on the refactoring!

hw_diag/tests/test_views_diagnostics.py
kashifpk: Once testing for hm-dashboard is available we could setup integration tests between hm-diag <-> HPT <-> hm-dashboard.

hw_diag/tests/test_views_diagnostics.py
```diff
@@ -86,3 +91,28 @@ def test_version_endpoint(self):
         self.assertEqual(reply['firmware_version'], '1337.13.37')
         self.assertEqual(reply['diagnostics_version'], 'aabbffe')
         self.assertEqual(reply['firmware_short_hash'], '0011223')
+
+    @patch('hw_diag.diagnostics.add_gateway_txn_diagnostic.AddGatewayTxnDiagnostic.perform_test')
+    def test_add_gateway_txn_success(self, mock):
+        endpoint = '/v1/add-gateway-txn'
+        data = dict(file=(BytesIO(), 'shipping_destination.json.asc'),)
+        resp = self.client.post(endpoint, content_type='multipart/form-data', data=data)
```
kashifpk, line 99: @posterzh - This seems wrong. Because the request would contain RAW POST data, not `multipart/form-data`.

hw_diag/utilities/miner.py
```diff
@@ -33,3 +34,8 @@ def fetch_miner_data(diagnostics):
     diagnostics['MN'] = peerbook.get('nat')
     diagnostics['MR'] = diagnostics['MN'] == 'symmetric'
     return diagnostics
+
+
+def create_add_gateway_txn(destination_wallet):
+    add_gateway_txn = client.create_add_gateway_txn(destination_wallet, NEBRA_WALLET_ADDRESS)
```
kashifpk, line 40: Not a blocker but the code at line 7 `client = MinerClient()` and similar pattern in other files should usually be defined inside the class/function that makes use of it. In this case, this line should have been inside both `fetch_miner_data` and `create_add_gateway_txn`.

The reasoning: Module level instances get created on module import and since they are module-level-global there are chances of more things going wrong with them. Like if one of the calling functions causes them to be in an invalid state then all other functions using them would also suffer.

One of the exceptions to this rule is singletons and entry points for web frameworks. Like the `app` object for Flask.

hw_diag/utilities/security.py
kashifpk: So in your current unit tests your test key gets added to your system/user-wide GPG store (in ~/.gnupg)? Do you think updating your actual system configs/settings is desirable just when you want to run some tests?

Like if I pull in these changes and run these tests at my end will they also the key to my GPG key store?

For the actual code calling `gnupg.GPG()` without homedir makes sense as we've already imported the key in Dockerfile. However for test cases it does not seem like a great option.

Also the same comment as above that gpg object should be created inside of the `get_verified_json(data)` function.

```python
def get_verified_json(data):
    # This decryption performs a signature validation and separate the payload from the rest of
    # the signature and not doing any real decryption because the message is ent in cleartext.
    gpg = gnupg.GPG()
    decrypted_data = gpg.decrypt(data)
```

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +114,46 @@ def version_information():
...
+            "fee": 65000,
+            "owner": "14GWyFj9FjLHzoN3aX7Tq7PL6fEg4dfWPY8CrK8b9S5ZrcKDz6S",
+            "payer": "138LbePH4r7hWPuTnK6HXVJ8ATM2QU71iVHzLTup1UbnPDvbxmr",
+            "staking fee": 4000000,
+            "txn": "CrkBCiEBrlImpYLbJ0z0hw5b4g9isRyPrgbXs9X+RrJ4pJJc9MkS..."
+        }
+    }
+    """
+
+    shipping_destination_with_signature = request.get_data()
```
kashifpk, line 139: Have you verified that this works with your test cases where data is being sent as multipart form data (normal HTTP Form POST)? AFAIK `get_data()` works when the POST data does not contain a mime-type/content-type. If we're passing `multipart/form-data' and content type this might cause get_data not to work as expected. At the very minimum based on the docs [here](https://tedboy.github.io/flask/generated/generated/flask.Request.get_data.html) we need to set `parse_form_data` to `True` like `request.get_data(parse_form_data=True)`.

I feel like we should double check and test it. Probably will do that later today when we do some manual integration tests.

hw_diag/views/diagnostics.py
```diff
@@ -111,3 +114,46 @@ def version_information():
...
+            DESTINATION_ADD_GATEWAY_TXN_KEY, err_msg)
+        return diagnostics_report, 406
+
+    diagnostics = [
+        AddGatewayTxnDiagnostic(shipping_destination_with_signature),
+    ]
+    diagnostics_report = DiagnosticsReport(diagnostics)
+    diagnostics_report.perform_diagnostics()
+    if diagnostics_report.has_errors({DESTINATION_WALLETS_KEY}):
+        http_code = 406
```
kashifpk, line 153: That's usually when data provided by the client has errors. In our case if the signed payload does not get "decrypted" correctly.

If there were errors during gateway transaction generation, isn't that a server side issue and more worthy of a 5xx error?

kashifpk (changes requested):
Probably need to double check some of the implementation.

hw_diag/utilities/security.py
posterzh: Make sense.

hw_diag/tests/test_views_diagnostics.py
posterzh: Good catch. Thank you.

hw_diag/views/diagnostics.py
posterzh: Got it. Let me update a bit and I will ensure a request without content type should work and also **the test case** should be without a content type.

hw_diag/views/diagnostics.py
posterzh: For the failure of add-gateway-txn, it can be considered as a server-side error, so I will return 500 in this case. Thank you.

Dockerfile
```diff
@@ -31,7 +31,16 @@ RUN \
...
         i2c-tools \
-        libdbus-1-3
+        libdbus-1-3 \
+        gpg
+
+# Nebra uses /opt by convention
+WORKDIR /opt/
+
+# Import gpg key
+COPY keys/kashif-manufacturing.pub ./
```
marvinmarnold, line 41: Can you change this back to prod key?

keys/kashif-manufacturing.pub
```diff
@@ -0,0 +1,24 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
marvinmarnold, line 1: Remove this

marvinmarnold (changes requested):
@posterzh after you remove Kashif's keys this looks good to me

Dockerfile
posterzh: yes, as planned I will revert it back.

keys/kashif-manufacturing.pub
posterzh: yes, as planned I will revert it back.

posterzh:
@marvinmarnold Now, Kashif's test key is removed and this branch is rebased onto master and resolved a conflict with `lte_diagnostic.py`.

marvinmarnold (approved):
@kashifpk can you conduct another review?

kashifpk:
> @kashifpk can you conduct another review?

@marvinmarnold - Sure, I'll run a couple of integration tests too.

MuratUrsavas (approved):

PR #308, opened 2022-03-01
feat: pull key storage bus from hardware def

shawaj:
**Issue**

- Link: https://github.com/NebraLtd/hm-pyhelper/pull/84
- Summary: ECC was hard coded

**How**

Instead of hard coding the key storage bus for ECC we will now make it pull in from pyhelper module

**Screenshots**

**References**

Relates-to: https://github.com/NebraLtd/hm-pyhelper/pull/84

Co-authored-by: radicale <radicale@users.noreply.github.com>

https://github.com/radicale/hm-diag/commit/a3fd21c22dc6230a32518dab347adb3a868e790d

**Checklist**

- [x] Tests added
- [x] Documentation added

### DIFF
--- a/hw_diag/utilities/hardware.py
+++ b/hw_diag/utilities/hardware.py
@@ -2,7 +2,7 @@ from time import sleep
 import dbus
 from hm_pyhelper.logger import get_logger
 from hm_pyhelper.miner_param import get_public_keys_rust
-from hm_pyhelper.hardware_definitions import variant_definitions, is_rockpi
+from hm_pyhelper.hardware_definitions import variant_definitions, get_variant_attribute
 from hw_diag.utilities.shell import config_search_param
 
 logging = get_logger(__name__)
@@ -186,14 +186,18 @@ def set_diagnostics_bt_lte(diagnostics):
 def detect_ecc(diagnostics):
     # The order of the values in the lists is important!
     # It determines which value will be available for which key
-    if is_rockpi():
-        commands = [
-            'i2cdetect -y 7'
-        ]
-    else:
-        commands = [
-            'i2cdetect -y 1'
-        ]
+    variant = diagnostics.get('VA')
+
+    try:
+        key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')
+    except Exception as e:
+        logging.warn("Unable to lookup storage bus from hardware definitions, "
+                     "falling back to the default. Exception message: {}".format(e))
+        key_storage_bus = '/dev/i2c-1'
+
+    commands = [
+            'i2cdetect -y ' + key_storage_bus.split('-')[1]
+    ]
 
     parameters = ["60 --"]
     keys = ["ECC"]
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@ requests==2.26.0
 retry==0.9.2
 sentry-sdk[Flask]==1.5.1
 dbus-python==1.2.16
-hm-pyhelper==0.13.10
+hm-pyhelper==0.13.13


hw_diag/utilities/hardware.py
```diff
@@ -186,14 +186,17 @@ def set_diagnostics_bt_lte(diagnostics):
...
-        commands = [
-            'i2cdetect -y 7'
-        ]
-    else:
-        commands = [
-            'i2cdetect -y 1'
-        ]
+    variant = diagnostics.get('VA')
+    variant_data = variant_definitions.get(variant)
+    if variant_data:
```
shawaj, line 191: @marvinmarnold is this the correct way to do this here for "DRY" purposes? Or do we already pull this info in somewhere already somewhere else in the diag module?

marvinmarnold: @shawaj [this](https://github.com/NebraLtd/hm-pyhelper/blob/06a752a196e05a602506db1b779a5fa62fd55f3d/hm_pyhelper/hardware_definitions.py#L439) helper method would be better.

shawaj: Like this @marvinmarnold :

```
def detect_ecc(diagnostics):
    # The order of the values in the lists is important!
    # It determines which value will be available for which key
    variant = diagnostics.get('VA')
    key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')

    if not key_storage_bus:
        key_storage_bus = '/dev/i2c-1'

    commands = [
            'i2cdetect -y ' + key_storage_bus.split('-')[1]
    ]

    parameters = ["60 --"]
    keys = ["ECC"]

    for (command, param, key) in zip(commands, parameters, keys):
        try:
            diagnostics[key] = config_search_param(command, param)
        except Exception as e:
            logging.error(e)
```

marvinmarnold: `get_variant_attribute` raises an exception instead of returning Falsey so it would be:

```
    variant = diagnostics.get('VA')

    try:
        key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')
    except KeyError as e:
        logging.warn("Unable to lookup storage bus from hardware definitions, falling back to the default.")
        key_storage_bus = '/dev/i2c-1'

    commands = [
            'i2cdetect -y ' + key_storage_bus.split('-')[1]
    ]
```

marvinmarnold: Whoops, it re-raises different exceptions. Do we want different behavior when variant unknown vs when variant doesn't have KEY_STORAGE_BUS set? Can either catch the [two](https://github.com/NebraLtd/hm-pyhelper/blob/06a752a196e05a602506db1b779a5fa62fd55f3d/hm_pyhelper/hardware_definitions.py#L448) exceptions [separately](https://github.com/NebraLtd/hm-pyhelper/blob/06a752a196e05a602506db1b779a5fa62fd55f3d/hm_pyhelper/hardware_definitions.py#L454) or use `except Exception as e`.

shawaj: I guess if it can't find the variant or there is no variant present as an ENV variable it should probably throw an error saying "unknown variant"

And if it is a known variant but missing the KEY_STORAGE_BUS then out should maybe say that instead.

However in both cases it should probably default to i2c-1 so that it at least attempts to find the ECC key (and i2c-1 is the most common of the options)

However, maybe we don't need a default at all and it's better in the case of a missing KEY_STORAGE_BUS to just provide an error so that it doesn't silently cause any other issues

Not sure what you think?

shawaj: TBH I think what you've put above but with `except Exception as e` is probably sufficient and just throw the message about falling back to the default:

```
    variant = diagnostics.get('VA')

    try:
        key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')
    except Exception as e:
        logging.warn("Unable to lookup storage bus from hardware definitions, falling back to the default.")
        key_storage_bus = '/dev/i2c-1'

    commands = [
            'i2cdetect -y ' + key_storage_bus.split('-')[1]
    ]
```

That seems reasonable to me

marvinmarnold (approved):

hw_diag/utilities/hardware.py
marvinmarnold: Ya that sounds fine to me. Also reminded me that we should eventually update [this test](https://github.com/NebraLtd/hm-pyhelper/blob/06a752a196e05a602506db1b779a5fa62fd55f3d/hm_pyhelper/tests/test_hardware_definitions.py#L15) to at least enforce `KEY_STORAGE_BUS` is present.

Issue #307, opened 2022-02-22
unknown error while downloading firmware: 'NoneType' object has no attribute 'get_property' (quectel)


vpetersson:
```
 diagnostics  2022-02-22 10:56:46,812 - [ERROR] - hw_diag.utilities.quectel - (quectel.py).ensure_quectel_health -- /opt/python-dependencies/hw_diag/utilities/quectel.py:(328) - unknown error while downloading firmware: 'NoneType' object has no attribute 'get_property'
 diagnostics  ERROR:hw_diag.utilities.quectel:unknown error while downloading firmware: 'NoneType' object has no attribute 'get_property'
 diagnostics  2022-02-22 10:56:46,832 - [INFO] - hw_diag.utilities.quectel - (quectel.py).ensure_quectel_health -- /opt/python-dependencies/hw_diag/utilities/quectel.py:(331) - skipping quectel firmware/settings update because UPDATE_QUECTEL_EG25G_MODEM is not set
 diagnostics  INFO:hw_diag.utilities.quectel:skipping quectel firmware/settings update because UPDATE_QUECTEL_EG25G_MODEM is not set
 diagnostics  INFO:apscheduler.executors.default:Job "get_app.<locals>.run_quectel_health_task (trigger: interval[1:00:00], next run at: 2022-02-22 11:54:43 UTC)" executed successfully
```

Looks like something went wrong with the update on [this device](https://dashboard.balena-cloud.com/devices/7525ea646143df1a2bfc063814121947/summary)

pritamghanghas:
has been fixed as part of #133.

Issue #133, opened 2021-08-26
AT&T 3G switch off - EG25-G firmware update


shawaj:
AT&T are switching off 3G in USA in February 2022 which means we need to apply a firmware update to any EG25G mPCIe modules for them to continue working on AT&T.

This can be done with AT commands or via a full firmware update. Full firmware update is preferable.

Need to work out the best way to roll this out.

- [Quectel_EC2x&EG9x&EG2x-G_Series_Software_Configuration_for_AT&T_3G_Sunset_Guide_V1.0.pdf](https://github.com/NebraLtd/hm-diag/files/7060740/Quectel_EC2x.EG9x.EG2x-G_Series_Software_Configuration_for_AT.T_3G_Sunset_Guide_V1.0.pdf)
- [AT&T 3G switch off document.docx](https://github.com/NebraLtd/hm-diag/files/7060741/AT.T.3G.switch.off.document.docx)
- [ATT 3G Sunset affected FW.pdf](https://github.com/NebraLtd/hm-diag/files/7060742/ATT.3G.Sunset.affected.FW.pdf)
- [Quectel_EOL 2021081201_EC25-EUX&EC21-EUX(Mini PCIe)_Module_EOL_Notification.pdf](https://github.com/NebraLtd/hm-diag/files/7060743/Quectel_EOL.2021081201_EC25-EUX.EC21-EUX.Mini.PCIe._Module_EOL_Notification.pdf)

- Check model and figure out if EG25-G is what we care about.
- Check firmware version.
- Include firmware update payload in container so its already there when needed.
- Figure out a method to updrade firmware
- Consider rollback logic after Xmin where connection hasn't established.

shawaj:
Just a ping on this @marvinmarnold @vpetersson as needs sorting by Feb

vpetersson:
@shawaj Do we have any of these boards so that we can test it?

shawaj:
yes we have thousands - it is just our standard mPCIe 4g module (think you have one in your outdoor miner?)

pritamghanghas:
### Plan to implement it:
# compiling QFireHose flashing tool into hm-diag docker image
* compile and install QFirehose as /usr/sbin/QFirehose. hm-diag repository carries the tool source and firmware in the following structure.

> quectel/
> ├── firmware
> │   └── eg25-g
> │       └── EG25GGBR07A08M2G_30.006.30.006
> │           ├── contents.xml
> │           ├── md5.txt
> │           └── update
[...]

PR #309, opened 2022-03-02
feat: add EG25G firmware and flashing tool (#133)

pritamghanghas:
* added lot more test cases
* fixed is_att_sim to handle exception on no modem found #307
* switch to sha256 based integrity check instead of md5

**133**

- Link: https://github.com/NebraLtd/hm-diag/issues/133
- Summary: mostly tests and small fixes here and there

**Checklist**

- [x] Tests added
- [x] Documentation added


MuratUrsavas (approved):
LGTM :+1:

hw_diag/tests/test_download.py
```diff
@@ -0,0 +1,119 @@
...
+        for filename in [OUTPUT_FILENAME, OUTPUT_FILENAME + ".part"]:
+            if os.path.exists(filename):
+                os.remove(filename)
+
+        responses.reset()
+
+    # @patch(requests.get, read_stream)
+    @patch("hw_diag.utilities.download.DEFAULT_BLOCK_SIZE", 1024)
+    @responses.activate
+    def test_download(self):
```
kashifpk, line 77: @pritamghanghas - Can you explain what this test is testing against? There are a lot of mocking and I'm not sure testing for download would make sense if we're actually mocking the download process?

pritamghanghas: testing that download function will save .part file if it doesn't receive full data. And subsequently finish it later on next call.

pritamghanghas: > @pritamghanghas - Can you explain what this test is testing against? There are a lot of mocking and I'm not sure testing for download would make sense if we're actually mocking the download process?

testing that download function will save .part file if it doesn't receive full data. And subsequently finish it later on next call.

kashifpk (commented):
@pritamghanghas - In general if there is a need for a lot of mocks in your test cases then using pytest fixtures becomes a better option to keep the tests more organized and readable. This is certainly not a blocker, just a side note for you to look into in the future.

I've mostly contained my view against python code and stuff and not the actual technical implementation. I think someone in the device team is in a better position to review that.

hw_diag/tests/test_quectel.py
```diff
@@ -1,47 +1,150 @@
...
+    @patch("hw_diag.utilities.quectel.MODEM_RESET_WAIT_TIME", 0.001)
+    @patch("hw_diag.utilities.quectel.INTERNET_MAX_WAIT_TIME", 0.001)
+    @patch("hw_diag.utilities.quectel.FW_STATE_FILE", MOCKED_STATE_FILE)
+    @patch("dbus.SystemBus")
+    @patch("dbus.Interface")
+    @patch(
+        "hw_diag.utilities.dbus_proxy.systemd_unit.SystemDUnit._wait_state",
+        return_value=True,
+    )
+    def test_firmware_upgrade_with_rollback(self, _, mock_interface, _2):
```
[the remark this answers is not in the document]
pritamghanghas: added comments. This ticket doesn't do much code change. Mostly tests and pep8/black changes.

pritamghanghas:
> 

will look up how fixtures work. Testing was a little difficult because most of the functionality was hardware specific.

hw_diag/utilities/quectel.py
```diff
@@ -18,50 +19,58 @@
...
+    "EG25GGBR07A08M2G_30.006.30.006.tgz":
+        "bde8796ae6307ecd31dce477d1a9281f67339ed0fbe4eee2d1ab108a061243b7",
 }
 
-GCP_FW_BASEPATH = ('https://storage.googleapis.com/helium-assets.nebra.com/'
-                   'modem-firmware/eg25g_firmware/')
-FW_STATE_FILE = '/var/data/quectel_state'
-FW_STORE_PATH = '/var/data/modem_firmware'
+GCP_FW_BASEPATH = (
+    "https://storage.googleapis.com/helium-assets.nebra.com/modem-firmware/eg25g_firmware/"
```
marvinmarnold, line 49: @shawaj @vpetersson this ultimately is used for the path to a `.tgz` with the firmware. There won't be nearly as much volume as the snapshots. I don't think it's necessary to switch this to nebracdn and may actually cause more confusion if they ask if this is "web traffic".

hw_diag/utilities/quectel.py
```diff
@@ -73,28 +82,35 @@ def find_eg25g_modem() -> Modem:
 @retry(Exception, tries=3, delay=5, backoff=2, logger=logging)
 def download_and_extract(url: str, file_path: str, file_hash: str) -> None:
     download_with_resume(url, file_path, file_hash)
-    untar_cmd = ['tar', '-xzf', file_path, '-C', FW_STORE_PATH]
-    tar_output = ''
+    untar_cmd = ["tar", "-xzf", file_path, "-C", FW_STORE_PATH]
```
marvinmarnold, line 85: @kashifpk what do you think about [tarfile](https://docs.python.org/3/library/tarfile.html) instead of shelling out?

@pritamghanghas not a blocker for me

hw_diag/utilities/quectel.py
```diff
@@ -73,28 +82,35 @@ def find_eg25g_modem() -> Modem:
...
+    tar_output = ""
     try:
         tar_output = subprocess.check_output(untar_cmd, stderr=subprocess.STDOUT)
     except Exception as e:
-        logging.error(f'Failed to untar {file_path}')
+        logging.error(f"Failed to untar {file_path}")
         raise e
     finally:
-        logging.debug(f'{untar_cmd} output: {tar_output}')
+        logging.debug(f"{untar_cmd} output: {tar_output}")
```
marvinmarnold, line 93: I would move this into the try, its not a very helpful command if the untar fails. That would also allow you to get rid of setting tar_output to empty string

hw_diag/utilities/quectel.py
```diff
@@ -150,121 +168,151 @@ def _do_upgrade(desired_fw_version: str) -> bool:
...
+    except Exception as e:
+        logging.info(f"failed to update setting {setter} to {desired_value}")
+        logging.error(f"error: {e}")
+        return False
 
-        old_value = getter_method()
-        if old_value == desired_value:
-            logging.info(f"Setting is already correct {old_value}")
+
+def update_setting_with_rollback(getter: str, setter: str, desired_value: str) -> bool:
```
marvinmarnold, line 243: `getter` and `setter` are confusing variable names. I associate them with instance functions but these are strings.

marvinmarnold: Ah OK, now I see how this is used in the code and it makes sense. It would be helpful to add documentation here that explains.

hw_diag/utilities/quectel.py
```diff
@@ -289,64 +338,68 @@ def ensure_modem_manager_health() -> None:
...
     we are limiting firmware upgrade to AT&T SIM for now because we have seen
     in testing that firmware update can effect non at&t connectivity like T-Mobile
     only gets ipv6 after firmware update which can create problems for us.
-    '''
+    """
     # make sure modem manager is responsive
     ensure_modem_manager_health()
 
     # we want to download firmware files at the first possible opportunity.
     try:
```
marvinmarnold, line 375: Isn't it a waste to download the firmware if not on ATT?

hw_diag/utilities/quectel.py
```diff
@@ -150,121 +168,151 @@ def _do_upgrade(desired_fw_version: str) -> bool:
...
+        logging.error(f"failed to set AT Setting: {e}")
         return False
     return True
 
 
 def firmware_upgrade_with_rollback() -> bool:
     modem = find_eg25g_modem()
     if not modem:
-        logging.debug('EG25G modem not found')
+        logging.debug("EG25G modem not found")
```
marvinmarnold, line 286: Does it even make sense to run the AT commands before this? I'd definitely wait to download until here.

marvinmarnold (changes requested):
Looking good. The only blocker is getting clarity about at what point we download the firmware and run the AT commands. I'm surprised we are doing either before detecting EG25 but maybe I missed something.

pritamghanghas:
> Looking good. The only blocker is getting clarity about at what point we download the firmware and run the AT commands. I'm surprised we are doing either before detecting EG25 but maybe I missed something.

All operations detect modem by calling find_eg25g_modem() on their own. We do this because lot of operations involve restarting modem and modem manager and modem might not be ready if they call operations blindly.

if modem is found, firmware will be downloaded for that modem and kept in local storage.

hw_diag/utilities/quectel.py
pritamghanghas: I think it is better if we have tar commands output logged in both cases success or failure.

hw_diag/utilities/quectel.py
pritamghanghas: We download and keep it as the plan is to upgrade firmware in all cases down the line. I also wanted to cover the case where a AT&T sim is installed in field.

hw_diag/utilities/quectel.py
pritamghanghas: AT commands as well as download checks for modem on their own. 
Reason: we start and stop modem multiple times to check for connectivity impact, modem might not be ready for dbus queries or might have changed its dbus object path.

hw_diag/utilities/quectel.py
pritamghanghas: added documentation

hw_diag/utilities/quectel.py
marvinmarnold: Won't tar_output always be empty string if `check_output` fails since a CalledProcessError will be raised.

marvinmarnold: You sure about that? If `check_output` raises an exception, it can't also return anything can it? From the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.check_output):

> If the return code was non-zero it raises a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError). The [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) object will have the return code in the [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError.returncode) attribute and any [output](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError.output) in the output attribute.

hw_diag/utilities/quectel.py
marvinmarnold: Yea I guess that is the most user friendly behavior. Is the download at least only happening if a modem is detected?

hw_diag/utilities/quectel.py
pritamghanghas: Done

marvinmarnold (commented):
LGTM now. Just want to give @shawaj and @vpetersson time to respond to: https://github.com/NebraLtd/hm-diag/pull/309/files#r819233511

hw_diag/utilities/quectel.py
kashifpk: Yes @marvinmarnold I'd second using tarfile module instead of shelling out. We all know how @vpetersson feels about shelling out instead of using available libs :-)

hw_diag/utilities/quectel.py
marvinmarnold: We will continue to use `helium-assets.nebra.com`

Issue #327, opened 2022-04-11
virtual env and caching in dockerfile


pritamghanghas:
balena python image goes out of its way to use the recent version of python which created a lot of issues installing grpcio wheel. It was becoming impossible to find a python wheel that matches balena image's glibc version and python 3.10. And building grpcio as part of 'pip install' take more than 1hr 40min. 

Using a virtual env with debian provided python is the only solution that worked. Addionally added caching support to our github docker build workflow.

**Acceptance Criterion**
- grpcio wheel gets used instead of setup.py
- docker file uses cache from container registry

marvinmarnold:
@pritamghanghas divide this up into multiple tickets. This sprint we will work on caching only.

pritamghanghas:
@marvinmarnold When I looked at it for division, I realized that all the original concerns have been fulfilled already.

PR #328, opened 2022-04-11
feat: virtual env and caching in dockerfile #327

pritamghanghas:
* use virtualenv with debian supplied python
* incorporate caching to and from dockerhub registry.

**[Issue](https://github.com/NebraLtd/hm-diag/issues/327)**

- Link: https://github.com/NebraLtd/hm-diag/issues/327
- Summary: add virtual env to have access to grpcio wheel and enable caching

**How**
- compile docker workflow was modified to push and pull buildcache from registry
- dockerfile was modified to:
  - make better use of build cache
  - use venv with debian supplied python version so that we can find grpcio wheel in pip

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,47 +3,59 @@
 ####################################################################################################
 ################################## Stage: builder ##################################################
 
-FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as builder
+FROM balenalib/raspberry-pi-debian:buster-build-20211014 as builder
 
-ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
+# Nebra uses /opt by convention
+WORKDIR /opt/
 
-RUN mkdir /tmp/build
-COPY ./ /tmp/build
-WORKDIR /tmp/build
-
-
-
+# this installs python 3.7.3 which is buster default and supported by grpcio wheel
 RUN \
     install_packages \
-            build-essential \
-            libdbus-glib-1-dev
+        python3-minimal \
+        python3-venv \
+        python3-pip \
+        build-essential \
+        libdbus-glib-1-dev
 
-RUN pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
+# This will be the path that venv uses for installation below
+ENV PATH="/opt/venv/bin:$PATH"
+
+# without pinning the pip version, we get linter warning
+COPY requirements.txt requirements.txt
+RUN python3 -m venv /opt/venv && \
+    pip install --no-cache-dir --upgrade pip==22.0.1 && \
+    pip install --no-cache-dir -r requirements.txt
 
 # firehose build, the tar is obtained from  quectel.
 # there is no install target in Makefile, doing manual copy
-RUN tar -xf quectel/qfirehose/QFirehose_Linux_Android_V1.4.9.tar
+COPY quectel quectel
+RUN tar -xf ./quectel/qfirehose/QFirehose_Linux_Android_V1.4.9.tar
 # docker linter wants WORKDIR for changing directory
-WORKDIR /tmp/build/QFirehose_Linux_Android_V1.4.9
+WORKDIR /opt/QFirehose_Linux_Android_V1.4.9
 RUN make && \
     cp QFirehose /usr/sbin/QFirehose && \
     rm -rf quectel/qfirehose
+
+WORKDIR /opt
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 Flask-APScheduler==1.12.2
 Flask-Caching==1.10.1
 Flask==2.0.1
+werkzeug==2.0.3
```
pritamghanghas, line 4: as dependency for flask version 2.1.0 was coming which was breaking tests due to 'as_tuple' argument removal.

requirements.txt
```diff
@@ -10,3 +11,4 @@ sentry-sdk[Flask]==1.5.1
 dbus-python==1.2.16
 hm-pyhelper==0.13.16
 python-gnupg==0.4.8
+grpcio==1.44.0
```
pritamghanghas, line 14: added so that it wheel installation can be tested.

Dockerfile
```diff
@@ -3,47 +3,60 @@
...
+
 # No need to cleanup the builder
 
 ####################################################################################################
 ################################### Stage: runner ##################################################
 
-FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner
-
-ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
+FROM balenalib/raspberry-pi-debian:buster-build-20211014 as runner
```
vpetersson, line 49: No, don't do this. This container is *a lot* larger.

pritamghanghas: > No, don't do this. This container is _a lot_ larger.

I can probably get rid of virtual env, but using balena's python creates problem with compiled wheel of grpcio.

vpetersson: We've had the conversation of  venv vs no venv. At the end of the day it doesn't really matter if you copy over `site-packages` or the `venv` environment from a size perspective.

With regards to the compilation -- i don't see why that matters if we cache the previous layer. It will just be skipped.

vpetersson: I want to keep the `balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner`, since this is shared across all other containers (and thus won't add any size for the base layer).

I'm not talking about pip caching, i'm talking about docker caching. You can read more [here](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei).

pritamghanghas: hm-config installs its own python, that can switch to balenalib/raspberry-pi-debian
hm-pktforwarder can also use the same. Doesn't have any specific dependency.

caching: I was also talking about docker caching, it is not working for source installed pip packages. Pip always compiles them and invalidate the layer cache.

vpetersson: That really shouldn't be the case. It should only invalidate when `requirements.txt` is altered 

https://stackoverflow.com/questions/48441296/pip-compiled-as-part-of-dockerfile-fastest-way-to-add-a-new-entry-to-requireme

pritamghanghas: the last "pip install . " which install hm-diag itself will mostly change because one of its source files would have changed. And then it will lookup dependencies -> which will be pyhelper -> which will depend on grpcio and that can only be installed with python setup.py because no wheel will be found matching python 3.10 and glibc-1.28. And we will be back to waiting 1hr 40 mins. 

There might be a way to preserve this cache. But each iteration I try on this will cost 1.4 hours. switching other to use balenalib/raspberry-pi-debian might be simpler solution IMHO. balenalib/raspberry-pi-debian-python will anyway likely be just one more layer on raspberry-pi-debian even if we don't switch.

vpetersson: Sure, but that is easily solved using a simple `requirements.txt` in conjunction with `setup.py` dependencies. That way by the time it gets to `pip install .`, it will already have all the dependencies installed from a previous layer where `requirements.txt` was installed.

pritamghanghas: that is already done in this pull request. The dockerfile was written to ensure that cache invalidation is a bare minimum. 

If using balenalib/raspberry-pi-debian-python instead of  balenalib/raspberry-pi-debian, that is acceptable. I don't think it will make a difference apart from having two python versions in the final container. But I am ok with that.

pritamghanghas: I have started using balena-debian-python again. Though I think if we have use a python image, we should push our own based on balena-debian.

PR #334, opened 2022-04-14
feat: robots.txt and headers

shawaj:
Add robots.txt

**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


shawaj:
@marvinmarnold @vpetersson shall we add the X-Robots-Tag to the add gateway and shutdown endpoints as well?

KevinWassermann94 (changes requested):
Linting needs to be fixed

shawaj:
> Linting needs to be fixed

Done

KevinWassermann94 (approved):

hw_diag/views/diagnostics.py
```diff
@@ -57,6 +57,7 @@ def read_diagnostics_file():
 def get_diagnostics_json():
     diagnostics = read_diagnostics_file()
     response = jsonify(diagnostics)
+    response.headers.set('X-Robots-Tag', 'none')
```
shawaj, line 60: @marvinmarnold can you set headers twice? Or will this overwrite the initial set?

shawaj: Maybe better way is to set like this https://www.adamsmith.haus/python/answers/how-to-set-response-headers-using-flask-in-python

KevinWassermann94: @shawaj It will override the existing headers. We should be able to uzse headers.append though

hw_diag/views/diagnostics.py
```diff
@@ -117,6 +122,8 @@ def version_information():
         'firmware_short_hash': os.getenv('FIRMWARE_SHORT_HASH', 'unknown'),
     }
 
+    response.headers.set('X-Robots-Tag', 'none')
```
shawaj, line 125: Seems like you can't set headers on a dictionary output?

vpetersson:
> @marvinmarnold @vpetersson shall we add the X-Robots-Tag to the add gateway and shutdown endpoints as well?

It's very improbable that those would get indexed given that we have no links to them anywhere.

shawaj:
> > @marvinmarnold @vpetersson shall we add the X-Robots-Tag to the add gateway and shutdown endpoints as well?
> 
> It's very improbable that those would get indexed given that we have no links to them anywhere.

Ah ok. Just was thinking for belt and braces. Looks like it won't work on version anyway so will just remove for now I guess

Issue #321, opened 2022-03-18
Light miner: Diagnostics page should NOT display blockchain details


marvinmarnold:
Miners no longer keep track of block height.

**Acceptance criteria**
- Miner container must be using hm-gatewayrs
- `Sync Percentage` removed
- `Miner Connected To Blockchain` removed
- `Height Status` renamed to `Validator height`
- Add `Validator block age`
- Add `Validator address`
- `Miner Relayed` removed from diagnostics page
- upnp removed from hm-config

shawaj:
Maybe this should be conditional initially? As we want to keep this until the full move

marvinmarnold:
@shawaj the [plan](https://github.com/NebraLtd/helium-miner-software/issues/399), we decided not to use feature flags and to instead work on a branch and plan to merge into production in discrete step.

marvinmarnold:
@pritamghanghas can you move this issue to the correct swim lane?

pritamghanghas:
Pausing work on this to work on light-hotspot QC [ticket](https://github.com/NebraLtd/hnt-products/issues/7). Will try to push the work in parallel if I find some time. 

Some potential issues for light-hotspot that will need some thought and work:
- We are using [rasp01.deb](https://github.com/NebraLtd/hm-gatewayrs/blob/main/Dockerfile#L13) from gateway-rs releases. This will work only on raspberry pi but not on rock pi. On Rockpi boards, the binary just hangs up. created a new [issue](https://github.com/NebraLtd/hm-gatewayrs/issues/30) for it.
- we are pulling gateway-mfr binary and putting it in the pypi package as a github action. I think it makes more sense to checkin this binary into our hm-pyhelper repo. In light hotspot we install from a git branch as we have not started publishing the light-spot branch and that leads to this binary being missing.

shawaj:
RasPi 1 Deb should work on RockPi too. Don't see why it wouldn't. Are you sure that's the reason the container is hanging?

pritamghanghas:
> RasPi 1 Deb should work on RockPi too. Don't see why it wouldn't. Are you sure that's the reason the container is hanging?
My bad on this one. While raising the issue, I confused it with gateway-mfr-rs. My issue was with gateway-mfr-rs. I will have to reconfirm which one was pulled for that.

PR #337, opened 2022-04-19
Light miner: Diagnostics page should NOT display local blockchain details #321

pritamghanghas:
**[#321](https://github.com/NebraLtd/hm-diag/issues/321)**

- Link: https://github.com/NebraLtd/hm-diag/issues/321
- Summary: remove local blockchain details and add validator blockchain details.

**How**
- Uses grpc to talk to locally running gateway-rs
- displays validator information in diag page. Shows "disconnected" when it can't connect to validator

**Screenshots**
diag_lighthotspot
diag_working

**References**

**Checklist**

- [x] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/diagnostics/gateway_diagnostics.py
+++ b/hw_diag/diagnostics/gateway_diagnostics.py
@@ -1,0 +1,105 @@
+from hm_pyhelper.diagnostics import DiagnosticsReport
+from hm_pyhelper.diagnostics.diagnostic import Diagnostic
+from hm_pyhelper.gateway_grpc.client import GatewayClient, decode_pub_key
+from hm_pyhelper.miner_param import LOGGER
+from hm_pyhelper.constants import diagnostics as DIAG_CONSTS
+import grpc
+
+
+class DiagnosticKeyInfo:
+    def __init__(self, friendly_name: str, short_key: str = None,
+                 grpc_method_name: str = None,
+                 composed_attribute_path: str = None):
+        """"
+        Represents a Gateway Diagnostic key required to fetch
+        corresponding value.
+        composed_attribute_path represents the path inside the value returned
+        by grpc call represented by grpc_method_name
+        if short_key is not supplied it defaults to friendly_name
+        """
+        self.friendly_name = friendly_name
+        self.short_key = short_key
+        if short_key is None:
+            self.short_key = friendly_name
+        self.grpc_method_name = grpc_method_name
+        self.composed_attribute_path = composed_attribute_path
+
+
+class GatewayDiagnostic(Diagnostic):
+    def __init__(self, keyinfo: DiagnosticKeyInfo):
+        super(GatewayDiagnostic, self).__init__(keyinfo.short_key, keyinfo.friendly_name)
+        self.keyinfo = keyinfo
+
+    def call_grpc(self, diagnostics_report: DiagnosticsReport,
+                  method_name: str, *args, **kwargs) -> object:
+        """
+        returns grpc return value if successful, None otherwise
+        """
+        try:
+            with GatewayClient() as client:
+                return getattr(client, method_name)(*args, **kwargs)
+        except grpc.RpcError as err:
+            LOGGER.error(f"rpc error: {err}")
+            LOGGER.exception(err)
+            diagnostics_report.record_failure(err, self)
+        except Exception as err:
+            LOGGER.exception(err)
+            diagnostics_report.record_failure(err, self)
+        return None
+
+    def perform_test(self, diagnostics_report):
+        ret_value = self.call_grpc(diagnostics_report,
+                                   self.keyinfo.grpc_method_name)
+
+        # if grpc has failed
+        if not ret_value:
+            return
+
+        # return simple value
+        if not self.keyinfo.composed_attribute_path:
+            diagnostics_report.record_result(ret_value, self)
+            return
+
+        # flatten the composite value
+        try:
+            for attribute_name in self.keyinfo.composed_attribute_path.split('.'):
+                ret_value = getattr(ret_value, attribute_name)
+            # this key requires additional decoding
+            if self.keyinfo.friendly_name == DIAG_CONSTS.VALIDATOR_ADDRESS_KEY:
+                ret_value = decode_pub_key(ret_value)
+            diagnostics_report.record_result(ret_value, self)
+        except Exception as err:
+            diagnostics_report.record_failure(err, self)
+
+
+class GatewayDiagnostics(Diagnostic):
+
+    SUPPORTED_GATEWAY_ATTRIBUTES = [
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.VALIDATOR_ADDRESS_KEY,
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='gateway.address'),
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.VALIDATOR_URI_KEY,
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='gateway.uri'),
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.VALIDATOR_BLOCK_AGE,
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='block_age'),
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.VALIDATOR_BLOCK_HEIGHT_KEY,
+                          short_key=DIAG_CONSTS.VALIDATOR_BLOCK_HEIGHT_SHORT_KEY,
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='height'),
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.GATEWAY_PUBKEY_KEY,
+                          grpc_method_name='get_pubkey'),
+        DiagnosticKeyInfo(friendly_name=DIAG_CONSTS.GATEWAY_REGION_KEY,
+                          short_key=DIAG_CONSTS.GATEWAY_REGION_SHORT_KEY,
+                          grpc_method_name='get_region'),
+    ]
+
+    def __init__(self):
+        self.gateway_diagnostics = []
+        for attribute in self.SUPPORTED_GATEWAY_ATTRIBUTES:
+            self.gateway_diagnostics.append(GatewayDiagnostic(attribute))
+
+    def perform_test(self, diagnostics_report: DiagnosticsReport) -> None:
+        for diagnostic in self.gateway_diagnostics:
+            diagnostic.perform_test(diagnostics_report)
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -4,17 +4,15 @@ import json
 
 from hm_pyhelper.hardware_definitions import variant_definitions
 from hm_pyhelper.miner_param import get_ethernet_addresses
-from hw_diag.utilities.blockchain import get_helium_blockchain_height
 from hw_diag.utilities.hardware import detect_ecc
 from hw_diag.utilities.hardware import get_serial_number
 from hw_diag.utilities.hardware import lora_module_test
 from hw_diag.utilities.hardware import set_diagnostics_bt_lte
 from hw_diag.utilities.hardware import get_public_keys_and_ignore_errors
-from hw_diag.utilities.miner import fetch_miner_data
 from hw_diag.utilities.shell import get_environment_var
 from hw_diag.utilities.gcs_shipper import upload_diagnostics
-from hm_pyhelper.miner_json_rpc.exceptions import MinerFailedFetchData
-from requests.exceptions import ConnectTimeout, ReadTimeout
+from hw_diag.diagnostics.gateway_diagnostics import GatewayDiagnostics
+from hm_pyhelper.diagnostics import DiagnosticsReport
 
 
 log = logging.getLogger()
@@ -24,7 +22,14 @@ log.setLevel(logging.DEBUG)
 def perform_hw_diagnostics(ship=False):  # noqa: C901
     log.info('Running periodic hardware diagnostics')
 
-    diagnostics = {}
+    diagnostics = [
+        GatewayDiagnostics()
+    ]
+
+    diagnostics_report = DiagnosticsReport(diagnostics)
+    diagnostics_report.perform_diagnostics()
+
+    diagnostics = diagnostics_report
 
     now = datetime.datetime.utcnow()
     diagnostics['last_updated'] = now.strftime("%H:%M UTC %d %b %Y")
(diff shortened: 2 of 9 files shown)


hw_diag/templates/diagnostics_page.html
```diff
@@ -102,27 +76,35 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
...
+                    <td class="text-success text-right">{{ diagnostics.validator_uri }} </td>
                   {% else %}
-                    <td class="text-warning text-right">False</td>
+                    <td class="text-warning text-right">Disconnected</td>
                   {% endif %}
                 </tr>
+                <tr>
+                    <td>Validator Block Height</td>
+                    {% if not diagnostics.MH %}
+                      <td class="text-warning text-right">Disconnected</td>
```
pritamghanghas, line 89: should we show errors as well. Or disconnected is enough.

hw_diag/templates/diagnostics_page.html
```diff
@@ -102,27 +76,35 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
...
+                  <td>Connected Validator</td>
+                  {% if diagnostics.validator_uri %}
+                    <td class="text-success text-right">{{ diagnostics.validator_uri }} </td>
                   {% else %}
-                    <td class="text-warning text-right">False</td>
+                    <td class="text-warning text-right">Disconnected</td>
                   {% endif %}
                 </tr>
+                <tr>
+                    <td>Validator Block Height</td>
```
pritamghanghas, line 87: open to better UI string for field names. These feel too verbose.

hw_diag/tests/test_get_miner_diag.py
```diff
@@ -1,78 +1,45 @@
...
-
 from unittest.mock import patch
-from copy import deepcopy
+import grpc
+from concurrent import futures
 
-import hm_pyhelper.miner_json_rpc.client
+import hm_pyhelper
+from hm_pyhelper.protos import local_pb2_grpc
+from hm_pyhelper.tests.test_gateway_grpc import TestData, MockServicer
```
pritamghanghas, line 8: Is it ok to reuse test data and code from hm_pyhelper? I can copy it but seemed like unncessary duplication.

requirements.txt
```diff
@@ -1,12 +1,14 @@
...
+werkzeug==2.0.3
 certifi==2021.5.30
 click==7.1.2
 gunicorn==20.1.0
 requests==2.26.0
 retry==0.9.2
 sentry-sdk[Flask]==1.5.1
 dbus-python==1.2.16
-hm-pyhelper==0.13.16
+hm-pyhelper @ https://github.com/NebraLtd/hm-pyhelper/archive/8013e7a.zip
```
pritamghanghas, line 13: this is a special commit that contains 32-bit gateway-mfr binary.

pritamghanghas:
the screenshot shows disconnected because gatewayrs is listening on localhost:4467. There doesn't seem to be a way to specify interface for api endpoint right now. Will ask helium and raise an issue. My feeling they might have already fixed it but gateway-rs hasn't had a release from quite a while.

hw_diag/tests/test_get_miner_diag.py
marvinmarnold: Seems reasonable to me.

hw_diag/templates/diagnostics_page.html
marvinmarnold: I'm OK with these but I get your point.

marvinmarnold (commented):
Looks pretty good. Please use more descriptive PR titles.

hw_diag/templates/diagnostics_page.html
```diff
@@ -102,27 +76,35 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                   <td class="text-right">{{ diagnostics.FRIENDLY }}</td>
                 </tr>
                 <tr>
-                  <td>Miner Connected To Blockchain</td>
-                  {% if diagnostics.MC %}
-                    <td class="text-success text-right">True</td>
+                  <td>Connected Validator</td>
+                  {% if diagnostics.validator_uri %}
```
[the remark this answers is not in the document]
pritamghanghas: from fetch_miner_data. It has changed, it is not getting all values from grpc-client.

hw_diag/templates/diagnostics_page.html
```diff
@@ -91,8 +65,8 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                 </tr>
                 <tr>
                   <td>Region Plan</td>
-                  {% if diagnostics.RE == 'UN123' %}
-                    <td class="text-right">Awaiting Location Assertion</td>
+                  {% if not diagnostics.RE %}
```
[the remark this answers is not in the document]
pritamghanghas: fetch_miner_data

pritamghanghas:
@marvinmarnold @vpetersson I want to merge this. it is reasonably tested. I can't run full test because of [this. ](https://github.com/helium/gateway-rs/issues/229). only workaround right now is to "do host networking for miner and diag". I don't want to do and test a change that we intend to revert.

hw_diag/utilities/miner.py
```diff
@@ -1,40 +1,30 @@
...
-        peerbook = client.get_peer_book()[0]
-        height = client.get_height()
-    except MinerFailedFetchData as err:
-        LOGGER.exception(err)
-        diagnostics['MC'] = "Failed to Fetch"
-        diagnostics['MD'] = "Failed to Fetch"
-        diagnostics['MH'] = None
-        diagnostics['MN'] = "Failed to Fetch"
-        diagnostics['MR'] = "Failed to Fetch"
+    with GatewayClient() as client:
```
marvinmarnold, line 8: Somehow I missed this when I first looked. Any code that we are changing in /json should be converted into a DiagnosticReport as done in /initFile.txt. This will require a bit of refactoring but will help with stability and consistency.

Extra context: https://github.com/NebraLtd/Hotspot-Production-Tool/issues/28

marvinmarnold (changes requested):

hw_diag/utilities/miner.py
pritamghanghas: I am not fully clear on this. Does it mean:
1. I segment perform_hw_diagnostics() functionality in to several different units of Diagnostic classe
2. and then modify /json to behave the same as /initFile

Shouldn't this be a separate ticket. I have only replaced the rpc client in this ticket.

pritamghanghas:
> @marvinmarnold @vpetersson I want to merge this. it is reasonably tested with mock servicer. Can't run full test on device because of [this. ](https://github.com/helium/gateway-rs/issues/229). only workaround right now is to "do host networking for miner and diag". Don't want to do and test a change that we intend to revert.

I am doing host networking. Based on the comments in the ticket, it doesn't look like helium guys will change their mind and allow binding to anything other than localhost.

hw_diag/utilities/miner.py
pritamghanghas: done

hw_diag/templates/diagnostics_page.html
```diff
@@ -92,38 +66,46 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
...
                 <tr>
                   <td>Variant</td>
                   <td class="text-right">{{ diagnostics.FRIENDLY }}</td>
                 </tr>
                 <tr>
-                  <td>Miner Connected To Blockchain</td>
-                  {% if diagnostics.MC %}
-                    <td class="text-success text-right">True</td>
+                  <td>Connected Validator</td>
+                  {% if diagnostics.gatewayrs.validator_uri %}
```
pritamghanghas, line 81: not checking for diagnostics.gatewayrs for validity. As far as I can see DiagnosticReport make sure that they key is there in case of failure also. 

Let me know if I am mistaken.

pritamghanghas (commented):
Fixed latest review comments.

hw_diag/diagnostics/gateway_diagnostics.py
```diff
@@ -0,0 +1,26 @@
+from hm_pyhelper.diagnostics import DiagnosticsReport
+from hm_pyhelper.diagnostics.diagnostic import Diagnostic
+from hw_diag.utilities.miner import fetch_miner_data
+from hm_pyhelper.miner_param import LOGGER
+import grpc
+
+
+class GatewayDiagnostic(Diagnostic):
+    KEY = 'GATEWAY'
```
marvinmarnold, line 9: There is no need to make KEY and FRIENDLY_NAME different. The reason we have two values is to be backwards compatible with logic that relies on old two letter keys like `E0` being present. For net-new entries that we add to the diagnostics report, we should make KEY and FRIENDLY_NAME the same.

hw_diag/diagnostics/gateway_diagnostics.py
```diff
@@ -0,0 +1,26 @@
...
+    FRIENDLY_NAME = "gatewayrs"
+
+    def __init__(self):
+        super(GatewayDiagnostic, self).__init__(self.KEY, self.FRIENDLY_NAME)
+
+    def perform_test(self, diagnostics_report: DiagnosticsReport) -> None:
+        diagnostics = {}
+        try:
+            fetch_miner_data(diagnostics)
```
marvinmarnold, line 18: It would be less confusing to read if this was `miner_data = fetch_miner_data()`

hw_diag/diagnostics/gateway_diagnostics.py
```diff
@@ -0,0 +1,26 @@
...
+        except grpc.RpcError as err:
+            LOGGER.error(f"rpc error: {err}")
+            LOGGER.exception(err)
+            diagnostics_report.record_failure(err, self)
+        except Exception as err:
+            LOGGER.exception(err)
+            diagnostics_report.record_failure(err, self)
+        diagnostics_report.record_result(diagnostics, self)
```
marvinmarnold, line 26: This pattern of scoping all of this data under the `gatewayrs` key makes perfect sense, but wasn't what I was expecting. I was expecting this [wrapper pattern](https://github.com/NebraLtd/hm-diag/blob/24240cc4088732536dacd129cdd99862f57158f3/hw_diag/diagnostics/key_diagnostics.py#L41) where several keys are added to the top level of the DiagnosticsReport. @kashifpk @NebraLtd/device-team what do you think is better?

hw_diag/tasks.py
```diff
@@ -1,30 +1,54 @@
...
-from requests.exceptions import ConnectTimeout, ReadTimeout
+from hw_diag.diagnostics.gateway_diagnostics import GatewayDiagnostic
+from hm_pyhelper.diagnostics import DiagnosticsReport
 
 
 log = logging.getLogger()
 log.setLevel(logging.DEBUG)
 
 
+def get_deep_key(source_dict: dict, key_path: list) -> Any:
```
marvinmarnold, line 24: Can you point me to where this method is used? Can't find it.

hw_diag/tasks.py
```diff
@@ -40,64 +64,18 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
...
-
     # Check the basics if they're fine and set an overall value
     # Basics are: ECC valid, Mac addresses aren't FF, BT Is present,
     # and LoRa hasn't failed
     if (
             diagnostics["ECC"] is True
             and diagnostics["E0"] is not None
             and diagnostics["W0"] is not None
-            and diagnostics["BT"] is True and diagnostics["LOR"] is True
+            and diagnostics["BT"] is True
```
marvinmarnold, line 76: Related to: https://github.com/NebraLtd/hm-config/issues/187#issuecomment-1094121136

@shawaj @KevinWassermann94 how do we want to treat PF on the diagnostics page in the immediate term?

hw_diag/templates/diagnostics_page.html
marvinmarnold: You probably want to use [check_errors](https://github.com/NebraLtd/hm-pyhelper/blob/b48f500b407d3c2ce8651844bb8a939746847691/hm_pyhelper/diagnostics/diagnostics_report.py#L49). If a Diagnostic calls record_result, then the key maps to a valid value. If it calls record_failure, then the key maps to an error value.

hw_diag/diagnostics/gateway_diagnostics.py
kashifpk: I'm ok with both approaches. Just a note that whichever we select should be uniform across all diagnostic reports so the code is easier to follow.

pritamghanghas: I find grouping simpler to read but it has one disadvantage. The group and child keys have to be checked separately for validity.

hw_diag/templates/diagnostics_page.html
pritamghanghas: Here I couldn't use check errors as render_template is passed json read from the file. It is not an instance of Diagnostic. But Will try and see if something else can be done. For now, It will show actual error that happened instead of "Disconnected". Disconnected will be shown only when the key itself is missing from json.

hw_diag/diagnostics/gateway_diagnostics.py
pritamghanghas: fixed.

hw_diag/diagnostics/gateway_diagnostics.py
pritamghanghas: this code no longer exists.

hw_diag/diagnostics/gateway_diagnostics.py
pritamghanghas: flattened it.

hw_diag/tasks.py
pritamghanghas: removed.

hw_diag/templates/diagnostics_page.html
```diff
@@ -92,38 +66,46 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                 </tr>
                 <tr>
                   <td>Region Plan</td>
-                  {% if diagnostics.RE == 'UN123' %}
-                    <td class="text-right">Awaiting Location Assertion</td>
+                  {% if diagnostics.miner_region %}
+                    <td class="text-right">{{ diagnostics.miner_region }}</td>
                   {% else %}
```
marvinmarnold, line 71: I would never expect the else to be reached, since miner_region will always have some value. Same goes for other diagnostics using the same pattern.

hw_diag/diagnostics/gateway_diagnostics.py
```diff
@@ -0,0 +1,104 @@
...
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='gateway.uri'),
+        DiagnosticKeyInfo(friendly_name='validator_block_age',
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='block_age'),
+        DiagnosticKeyInfo(friendly_name='validator_height',
+                          short_key='MH',
+                          grpc_method_name='get_validator_info',
+                          composed_attribute_path='height'),
+        DiagnosticKeyInfo(friendly_name='miner_pubkey',
```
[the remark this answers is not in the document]
pritamghanghas: as mentioned [here](https://github.com/NebraLtd/hm-diag/pull/337#discussion_r858668223) I can't do check errors because only plain dict is exposed to the template. I am trying creating DiagnosticReport out of json and expose that. Will update PR if that works.

hw_diag/templates/diagnostics_page.html
```diff
@@ -92,38 +66,46 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                 </tr>
                 <tr>
                   <td>Region Plan</td>
-                  {% if diagnostics.RE == 'UN123' %}
-                    <td class="text-right">Awaiting Location Assertion</td>
+                  {% if diagnostics.miner_region %}
```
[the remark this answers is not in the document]
pritamghanghas: as mentioned [here](https://github.com/NebraLtd/hm-diag/pull/337#discussion_r858668223) I can't do check errors because only plain dict is exposed to the template. I am trying creating DiagnosticReport out of json and expose that. Will update PR if that works.

requirements.txt
```diff
@@ -1,13 +1,15 @@
...
+werkzeug==2.0.3
 certifi==2021.5.30
 click==7.1.2
 gunicorn==20.1.0
 requests==2.26.0
 retry==0.9.2
 sentry-sdk[Flask]==1.5.1
 dbus-python==1.2.16
-hm-pyhelper==0.13.16
+hm-pyhelper @ https://github.com/NebraLtd/hm-pyhelper/archive/fc9abe4.zip
```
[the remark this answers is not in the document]
pritamghanghas: pyhelper also has a branch for light-miner and which breaks json rpc. Do we want to do release from light-miner?

hw_diag/templates/diagnostics_page.html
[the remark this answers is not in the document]
pritamghanghas: now DiagnosticReport is exposed instead of plain dict. There doesn't seem to be any regression because of that and it will correctly show disconnected.

hw_diag/templates/diagnostics_page.html
pritamghanghas: fixed by exposing DiagnosticsReport instead of simple dict.

hw_diag/diagnostics/gateway_diagnostics.py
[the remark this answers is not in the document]
pritamghanghas: changed

requirements.txt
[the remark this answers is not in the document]
marvinmarnold: Makes sense. Tricky. This does seem fine for now.

hw_diag/diagnostics/gateway_diagnostics.py
```diff
@@ -0,0 +1,104 @@
...
+                ret_value = decode_pub_key(ret_value)
+            diagnostics_report.record_result(ret_value, self)
+        except Exception as err:
+            diagnostics_report.record_failure(err, self)
+
+
+class GatewayDiagnostics(Diagnostic):
+
+    SUPPORTED_GATEWAY_ATTRIBUTES = [
+        DiagnosticKeyInfo(friendly_name='validator_address',
```
marvinmarnold, line 77: I just remembered that we want to [define these keys in pyhelper](https://github.com/NebraLtd/hm-pyhelper/tree/master/hm_pyhelper/constants). That way other services or HPT can import pyhelper and use those constants instead of checking for hardcoded strings.

pritamghanghas: Done.

marvinmarnold (approved):

PR #339, opened 2022-04-24
Display CPU Temperature

gfelisberto:
**Issue**

- Link:
- Summary:

Fetch the CPU temperature to the diagnostics object and also display in the diagnostics page.

### DIFF
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -3,6 +3,7 @@ import datetime
 import json
 
 from hm_pyhelper.hardware_definitions import variant_definitions
+from hm_pyhelper.miner_param import get_temperatures
 from hm_pyhelper.miner_param import get_ethernet_addresses
 from hw_diag.utilities.blockchain import get_helium_blockchain_height
 from hw_diag.utilities.hardware import detect_ecc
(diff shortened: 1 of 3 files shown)


hw_diag/tasks.py
```diff
@@ -30,6 +31,7 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
     diagnostics['last_updated'] = now.strftime("%H:%M UTC %d %b %Y")
 
     get_ethernet_addresses(diagnostics)
+    get_temperatures(diagnostics)
```
marvinmarnold, line 34: Can you rebase on https://github.com/NebraLtd/hm-diag/pull/337 and convert this into a Diagnostic?

hw_diag/templates/diagnostics_page.html
```diff
@@ -148,6 +148,14 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                     <td class="text-danger text-right">N/A</td>
                   {% endif %}
                 </tr>
+                <tr>
+                  <td>CPU Temperature</td>
+                  {% if diagnostics.CPU0 %}
+                  <td class="text-right">{{ diagnostics.E0 }} °C</td>
```
marvinmarnold, line 154: I think you meant `CPU0`. It should also be further indented.

marvinmarnold (changes requested):
Thanks for your contribution. Some changes requested.

hw_diag/tasks.py
marvinmarnold: Sorry, rebasing on #337 doesn't make sense. But please look at the DiagnosticReport pattern that is being introduced there and copy here.

gfelisberto: I have almost zero Python skills and could not follow the process

PR #340, opened 2022-04-25
feat: virtual env and caching in dockerfile #327

pritamghanghas:
* use virtualenv with debian supplied python
* incorporate caching to and from dockerhub registry.

**[Issue](https://github.com/NebraLtd/hm-diag/issues/327)**

- Link: https://github.com/NebraLtd/hm-diag/issues/327
- Summary: add virtual env to have access to grpcio wheel and enable caching

**How**
- compile docker workflow was modified to push and pull buildcache from registry
- dockerfile was modified to:
  - make better use of build cache
  - use venv with debian supplied python version so that we can find grpcio wheel in pip

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/.github/workflows/compile-docker.yml
+++ b/.github/workflows/compile-docker.yml
@@ -48,6 +48,8 @@ jobs:
           push: true
           labels: ${{ steps.meta.outputs.labels }}
           tags: ${{ steps.meta.outputs.tags }}
+          cache-from: type=registry,ref=nebraltd/hm-diag:buildcache
+          cache-to: type=registry,ref=nebraltd/hm-diag:buildcache,mode=max
 
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,47 +3,59 @@
 ####################################################################################################
 ################################## Stage: builder ##################################################
 
-FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as builder
+FROM balenalib/raspberry-pi-debian:buster-build-20211014 as builder
 
-ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
+# Nebra uses /opt by convention
+WORKDIR /opt/
 
-RUN mkdir /tmp/build
-COPY ./ /tmp/build
-WORKDIR /tmp/build
-
-
-
+# this installs python 3.7.3 which is buster default and supported by grpcio wheel
 RUN \
     install_packages \
-            build-essential \
-            libdbus-glib-1-dev
+        python3-minimal \
+        python3-venv \
+        python3-pip \
+        build-essential \
+        libdbus-glib-1-dev
 
-RUN pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
+# This will be the path that venv uses for installation below
+ENV PATH="/opt/venv/bin:$PATH"
+
+# without pinning the pip version, we get linter warning
+COPY requirements.txt requirements.txt
+RUN python3 -m venv /opt/venv && \
+    pip install --no-cache-dir --upgrade pip==22.0.1 && \
+    pip install --no-cache-dir -r requirements.txt
 
 # firehose build, the tar is obtained from  quectel.
 # there is no install target in Makefile, doing manual copy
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -3,47 +3,59 @@
...
 ################################### Stage: runner ##################################################
 
-FROM balenalib/raspberry-pi-debian-python:buster-run-20211014 as runner
-
-ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
+FROM balenalib/raspberry-pi-debian-python:buster-build-20211014 as runner
 
 RUN \
     install_packages \
+        python3-minimal \
```
marvinmarnold, line 52: Surprising to see python packages installed because:
- Already using Python Buster image: raspberry-pi-debian-python.
- Using venv, so Python should be bundled with that.

.github/workflows/compile-docker.yml
```diff
@@ -48,6 +48,8 @@ jobs:
           push: true
           labels: ${{ steps.meta.outputs.labels }}
           tags: ${{ steps.meta.outputs.tags }}
+          cache-from: type=registry,ref=nebraltd/hm-diag:buildcache
```
marvinmarnold, line 51: What work still needs to be done [here](https://github.com/NebraLtd/product-management/issues/53) given this change?

requirements.txt
```diff
@@ -10,3 +11,4 @@ sentry-sdk[Flask]==1.5.1
 dbus-python==1.2.16
 hm-pyhelper==0.13.16
 python-gnupg==0.4.8
+grpcio==1.44.0
```
marvinmarnold, line 14: This probably belongs as part of a different PR. it's not related to Docker caching.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 Flask-APScheduler==1.12.2
 Flask-Caching==1.10.1
 Flask==2.0.1
+werkzeug==2.0.3
```
marvinmarnold, line 4: Why is this needed now?

Dockerfile
pritamghanghas: I agree with venv pulling python, I can remove that I think without any issues.

raspberrypi-debian-python image is 3.10 that creates problem for precompiled grpcio wheel. Long discussion in the abondoned [pull request](https://github.com/NebraLtd/hm-diag/pull/328)

TDLR:
* balena packages their own compiled python 3.10 linked against buster's glibc2.28
* grpcio wheels that target python 3.10, target glibc2.29. Which is generally the combo in distributions that carrry 3.10
* To use those wheels, we need python 3.7 to 3.9 which is what buster shipped. So this install
* If I don't do that, even after caching grpc compilation creates slow docker builds. 
* runner still uses the base image as buster-python because of discussion in the old PR about keeping same runner base image.

.github/workflows/compile-docker.yml
pritamghanghas: Similar change will have to be done for other repositories.

requirements.txt
pritamghanghas: Yes it was there for testing that wheels work with the venv change.

requirements.txt
pritamghanghas: will have to check by removing it. It might just be an artifact of taking changes from integration branch where I was testing it.

Dockerfile
vpetersson: We could always compile and cache the wheels in GitHub

pritamghanghas: yes, that can be an option. But there will be some work there it might be easier to push nebra-raspberrypi-buster-python with upstream python.

marvinmarnold:
@pritamghanghas you aren't blocked by review here are you? Looks like you are going to follow up on some of the comments?

pritamghanghas:
@marvinmarnold No, not blocked. I will work on splitting it.

requirements.txt
pritamghanghas: removed it.

requirements.txt
pritamghanghas: removed it

pritamghanghas:
Didn't split it because now all the layer changes are only in builder. I think most of the original objections don't matter as runner is almost same.

marvinmarnold (approved):

Issue #357, opened 2022-05-10
Consolidate diagnostics page changes for light hotspot activation


marvinmarnold:
**Acceptance criteria**
- Remove: sync percentage, miner relayed, height, connected to blockchain
- Add: Validators if possible

PR #358, opened 2022-05-11
Consolidate diagnostics page changes for light hotspot activation #357

pritamghanghas:
**[Issue](https://github.com/NebraLtd/hm-diag/issues/357)**

- Link: https://github.com/NebraLtd/hm-diag/issues/357
- Summary: provide option to hide blockchain infrmation.

**How**
* introduced DISPLAY_MINER_INFO env vairable. It can be set to false to
hide display of miner info on diagnostic page.
* defaults to true

**Screenshots**
With Miner Info disabled:
light_page

With Miner Info enabled:
full_miner_page

**References**

**Checklist**

- [ ] Tests added
- [x] Documentation added

### DIFF
--- a/hw_diag/templates/diagnostics_page_light_miner.html
+++ b/hw_diag/templates/diagnostics_page_light_miner.html
@@ -1,0 +1,156 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta http-equiv="refresh" content="300">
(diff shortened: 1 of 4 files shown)


hw_diag/templates/diagnostics_page_light_miner.html
```diff
@@ -0,0 +1,156 @@
+<!doctype html>
```
pritamghanghas, line 1: added a new copy of template because layout has to be modified to make it symmetric. Trying to do layout with lot of if else was becoming unreadable.

PR #373, opened 2022-05-17
Handle empty responses from Balena Supervisor API

kashifpk:
Better handling of empty json response from balena supervisor API and additional test cases for testing this outcome.

The root cause of the shutdown logic failing was that we were not raising an exception (required by ShutdownGatewayDiagnosticReport in order to register diagnostic failure). So even though we logged the error it looked like success in the diagnostic logic.

**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [x] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/utilities/balena_supervisor.py
+++ b/hw_diag/utilities/balena_supervisor.py
@@ -1,4 +1,5 @@
 import os
+
 import requests
 from requests.exceptions import ConnectionError, ConnectTimeout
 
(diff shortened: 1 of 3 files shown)


pritamghanghas (approved):

hw_diag/utilities/balena_supervisor.py
```diff
@@ -43,17 +44,23 @@ def shutdown(self):
...
-        LOGGER.info(response.json())
         if response is None or response.ok is False:
             LOGGER.error("Device status request failed.")
             raise Exception("Device status request failed")
 
-        return response.json()[key_to_return]
+        try:
+            return response.json()[key_to_return]
+        except Exception:
+            return 'Failed due to supervisor API issue'
```
marvinmarnold, line 66: @kashifpk why return a string here instead of raising an exception?

kashifpk: > @kashifpk why return a string here instead of raising an exception?

@marvinmarnold - That's a mistake at my end thinking that we'll just display this message as device status as the return value is expected to be a string. However this won't cause the device diagnostic to fail which is incorrect behavior. I'll open a new PR and correct this.

kashifpk:
New PR is here: https://github.com/NebraLtd/hm-diag/pull/374

PR #374, opened 2022-05-18
Raise proper exeception for failed device status

kashifpk:
Instead of returning a string as a failed device status properly raise exception so the diagnostic fails correctly.

**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [x] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/utilities/balena_supervisor.py
+++ b/hw_diag/utilities/balena_supervisor.py
@@ -63,4 +63,8 @@ class BalenaSupervisor:
         try:
             return response.json()[key_to_return]
         except Exception:
+            LOGGER.error(
+                f"Couldn't find {key_to_return} key in response.\n"
+                f"Response content: {response.content}"
+            )
             raise RuntimeError('Supervisor API did not return valid json response')
--- a/hw_diag/utilities/balena_supervisor.py
+++ b/hw_diag/utilities/balena_supervisor.py
@@ -58,9 +58,9 @@ class BalenaSupervisor:
         response = self._make_request('GET', '/v2/state/status')
         if response is None or response.ok is False:
             LOGGER.error("Device status request failed.")
-            raise Exception("Device status request failed")
+            raise RuntimeError("Device status request failed")
 
         try:
             return response.json()[key_to_return]
         except Exception:
-            return 'Failed due to supervisor API issue'
+            raise RuntimeError('Supervisor API did not return valid json response')
--- a/hw_diag/utilities/balena_supervisor.py
+++ b/hw_diag/utilities/balena_supervisor.py
@@ -56,15 +56,24 @@ class BalenaSupervisor:
         LOGGER.info("Retrieving device status using Balena supervisor.")
 
         response = self._make_request('GET', '/v2/state/status')
-        if response is None or response.ok is False:
-            LOGGER.error("Device status request failed.")
-            raise RuntimeError("Device status request failed")
+
+        error_msg = ""
+        if response is None:
+            error_msg = "Device status request failed. No response recieved."
+
+        elif response.ok is False:
+            error_msg = "Device status request failed. Got non-OK response: " + \
+                  f"{response.status_code} {response.content}"
+
+        if error_msg:
+            LOGGER.error(error_msg)
+            raise RuntimeError(error_msg)
 
         try:
             return response.json()[key_to_return]
         except Exception:
-            LOGGER.error(
-                f"Couldn't find {key_to_return} key in response.\n"
-                f"Response content: {response.content}"
-            )
-            raise RuntimeError('Supervisor API did not return valid json response')
+            error_msg = "Supervisor API did not return valid json response.\n" + \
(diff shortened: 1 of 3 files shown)


marvinmarnold (approved):

marvinmarnold (changes requested):
In the future, I think we should also return the entire response to HPT via the DiagnosticReport so that it can be logged in cases where the miner does not have Balena access to report it's logs directly. May help debugging in the future.

hw_diag/utilities/balena_supervisor.py
```diff
@@ -58,9 +58,9 @@ def get_device_status(self, key_to_return) -> str:
         response = self._make_request('GET', '/v2/state/status')
         if response is None or response.ok is False:
             LOGGER.error("Device status request failed.")
-            raise Exception("Device status request failed")
+            raise RuntimeError("Device status request failed")
 
         try:
             return response.json()[key_to_return]
```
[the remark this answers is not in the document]
kashifpk: Done.

kashifpk:
I think I've logged it in the wrong place for now. We wanted the response content to be present in diagnostic response (`initFile.txt` payload) so we can log it from hpt. I'll fix that and then re-open the PR for review.

marvinmarnold:
Just to be clear, this what I was expecting for now. Additional logging in initFile.txt is a nice-to-have but I don't personally consider it a blocker. But great if the work is done.

kashifpk:
> Just to be clear, this what I was expecting for now. Additional logging in initFile.txt is a nice-to-have but I don't personally consider it a blocker. But great if the work is done.

Yes but that does not give us visibility of response inside diagnostics report errors from HPT  as part of initFile.txt payload. And now this is done too so we're covered both ways.

kashifpk:
@marvinmarnold - please re-review.

hw_diag/utilities/balena_supervisor.py
```diff
@@ -56,11 +56,24 @@ def get_device_status(self, key_to_return) -> str:
...
         response = self._make_request('GET', '/v2/state/status')
-        if response is None or response.ok is False:
-            LOGGER.error("Device status request failed.")
-            raise Exception("Device status request failed")
+
+        error_msg = ""
+        if response is None:
+            error_msg = "Device status request failed. No response recieved."
+
+        elif response.ok is False:
```
marvinmarnold, line 64: nit: `elif not response.ok` would be more idiomatic

hw_diag/utilities/balena_supervisor.py
```diff
@@ -56,11 +56,24 @@ def get_device_status(self, key_to_return) -> str:
         LOGGER.info("Retrieving device status using Balena supervisor.")
 
         response = self._make_request('GET', '/v2/state/status')
-        if response is None or response.ok is False:
-            LOGGER.error("Device status request failed.")
-            raise Exception("Device status request failed")
+
+        error_msg = ""
```
marvinmarnold, line 60: nit: False or None might be more clear

hw_diag/utilities/balena_supervisor.py
```diff
@@ -56,11 +56,24 @@ def get_device_status(self, key_to_return) -> str:
...
+        if response is None:
+            error_msg = "Device status request failed. No response recieved."
+
+        elif response.ok is False:
+            error_msg = "Device status request failed. Got non-OK response: " + \
+                  f"{response.status_code} {response.content}"
+
+        if error_msg:
+            LOGGER.error(error_msg)
```
marvinmarnold, line 69: Should this be warning level? We expect invalid responses occasionally. We are in the process of making Sentry only listen for CRITICAL level logs but when we eventually switch back to ERROR level, this shouldn't trigger Sentry.

marvinmarnold (approved):
Some nuanced comments, but looks good.

hw_diag/utilities/balena_supervisor.py
kashifpk: A matter of preference but per zen of python "explicit is better than implicit" and I've seen some code linting tools like python-black suggest comparisons like `is False` instead of `not value`.

hw_diag/utilities/balena_supervisor.py
kashifpk: Makes sense. Unfortunately I merged the branch before seeing this comment. Have updated it and pushed the branch a fresh now.

hw_diag/utilities/balena_supervisor.py
kashifpk: yeah but this empty string is usually set as start value so if we need to concatenate multiple  error messages into a string later it becomes easier instead of needing to convert None of False to str first.

hw_diag/utilities/balena_supervisor.py
kashifpk: @marvinmarnold - the new PR with this change is [here](https://github.com/NebraLtd/hm-diag/pull/375)

Issue #388, opened 2022-06-09
Remove freq from diagnostics


KevinWassermann94:
Acceptance Criteria
- [ ] Create new diagnostics to read region from file and show frequency based off of that

pritamghanghas:
@KevinWassermann94 Shall we totally remove this. I don't know what additional value it provides when "region" key is already there and it is displayed under "Region Plan" in diagnostic page. Seems redundant.

KevinWassermann94:
Makes sense. The only difference is that the region plan will be unavailable when no location has been asserted yet.
@marvinmarnold What do you think?

pritamghanghas:
@KevinWassermann94 Frequency will also be unavailable when unasserted. Our plan was is to deduce it from region and that is how miner decides it.

KevinWassermann94:
You are right

pritamghanghas:
blocked by https://github.com/NebraLtd/hotspot-production-images/issues/49
once frequency file injection is done.

marvinmarnold:
@NebraLtd/device-team can we remove this from the sprint? I don't think we want to do this yet.

marvinmarnold:
@NebraLtd/device-team this has been removed from the sprint

PR #389, opened 2022-06-10
feat: Remove freq from diagnostics #388

pritamghanghas:
* remove freq environment variable

**[Issue](https://github.com/NebraLtd/hm-diag/issues/388)**

- Link: https://github.com/NebraLtd/hm-diag/issues/388
- Summary: remove freq from environment variables.

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -90,6 +90,10 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         # No region found, put a dummy region in
         diagnostics['RE'] = "UN123"
 
+    # add frequency keys
+    diagnostics['FREQ'] = diagnostics['RE']  # friendly key
+    diagnostics['FR'] = diagnostics['RE']  # old key
+
     # Check the basics if they're fine and set an overall value
     # Basics are: ECC valid, Mac addresses aren't FF, BT Is present,
     # and LoRa hasn't failed
--- a/hw_diag/tasks.py
+++ b/hw_diag/tasks.py
@@ -90,10 +90,6 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         # No region found, put a dummy region in
         diagnostics['RE'] = "UN123"
 
-    # add frequency keys
-    diagnostics['FREQ'] = diagnostics['RE']  # friendly key
-    diagnostics['FR'] = diagnostics['RE']  # old key
-
     # Check the basics if they're fine and set an overall value
     # Basics are: ECC valid, Mac addresses aren't FF, BT Is present,
     # and LoRa hasn't failed
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -88,7 +88,11 @@
                 </tr>
                 <tr>
                   <td>Frequency</td>
-                  <td class="text-right">{{ diagnostics.FR }}</td>
+                  {% if diagnostics.FREQ %}
+                    <td class="text-right">{{ diagnostics.FREQ }}</td>
+                  {% else %}
+                    <td class="text-right">Requires location assertion</td>
+                  {% endif %}
                 </tr>
                 <tr>
                   <td>Region Plan</td>
--- a/hw_diag/templates/diagnostics_page.html
+++ b/hw_diag/templates/diagnostics_page.html
@@ -87,14 +87,6 @@
                   <td class="text-right">{{ diagnostics.FW }} ({{ diagnostics.firmware_short_hash }})</td>
                 </tr>
                 <tr>
-                  <td>Frequency</td>
-                  {% if diagnostics.FREQ %}
-                    <td class="text-right">{{ diagnostics.FREQ }}</td>
-                  {% else %}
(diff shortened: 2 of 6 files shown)


hw_diag/templates/diagnostics_page.html
```diff
@@ -88,7 +88,11 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                 </tr>
                 <tr>
                   <td>Frequency</td>
-                  <td class="text-right">{{ diagnostics.FR }}</td>
+                  {% if diagnostics.FREQ %}
```
pritamghanghas, line 91: shall we completely remove this. We already have "Region Plan" which will have same value.

hw_diag/tasks.py
```diff
@@ -90,6 +90,10 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
         # No region found, put a dummy region in
         diagnostics['RE'] = "UN123"
 
+    # add frequency keys
+    diagnostics['FREQ'] = diagnostics['RE']  # friendly key
```
pritamghanghas, line 94: didn't do a diagnostic for the time being as this already comes from Gateway_diagnostic in light_gateway branch. We will merge that soon.

hw_diag/templates/diagnostics_page.html
MuratUrsavas: Came to say the same. We should remove it completely. Otherwise, looks good to me.

posterzh (approved):
LGTM

MuratUrsavas (approved):

vpetersson:
Hmm that's a very good point. We do need frequency for labels. I guess we need to inject this into the disk image somehow. Maybe just a file on disk?

marvinmarnold:
I can't remember the specific blockers, but we weren't able to get this working when we went down the [nebra.json](https://github.com/NebraLtd/hotspot-production-images/pull/27) path. I believe the issue boils down to:

- Not able to mount a volume on a container to the file on the host
- Cannot inject this information into config.json

Related to: https://github.com/NebraLtd/hotspot-production-images/pull/27

Trying to figure out how to inject this information will eat up a lot of time. I think we should either:

1. Keep frequency fleets
2. Require manufacturing to input the frequency as part of running HPT

#2 would be easy but likely to result in mislabeling. #1 is safer and my vote. @NebraLtd/device-team

kashifpk:
I'll vote for option 1 too, option 2 is just asking for trouble.

vpetersson:
@marvinmarnold 

> I can't remember the specific blockers, but we weren't able to get this working when we went down the https://github.com/NebraLtd/hotspot-production-images/pull/27 path. I believe the issue boils down to:

IIRC it was the issue that we couldn't properly access it from user space.

> Trying to figure out how to inject this information will eat up a lot of time. I think we should either:

I don't like either of these tbh.

1 will lead to a lot of miss prints and 2 will create even more fleets than we have today (i.e. the opposite of what we want).

The issue we had when trying to do `nebra.json` for manufacturing was that we wanted read/write access to it. If we just need read access on first boot (essentially), that gives us more options. Maybe there's another path that we can write to? Remember that one of the per-requisites for `nebra.json` was that it should be both persistent (i.e. survive OS updates) and have read/write access. In this case, we don't really care about persistence nor read/write. As such, maybe writing it to `/etc` or say `/opt` will suffice.

pritamghanghas:
@marvinmarnold @kashifpk I don't know the full flow of manufacturing yet. But It seems we are already asking them to differentiate devices while flashing. After this we will ask them to differentiate while configuring. Does it cause lot of trouble with manufacturing.

MuratUrsavas:
Open up, manufacturing-doctor here!

I think we're approaching to the problem from a wrong angle. The important part is, we have a fixed frequency concentrator and don't have a chance to read the freq automatically as it is only changed with passive components without and permanent digital info on the concentrator chip. (We actually can do something about it, but let's assume we can't)

So in such manufacturing condition, a permanent information would be created at manufacturing time. And later that information would be propagated wherever needed. Fleets and devices pages are such examples. They do get information from manufacturing. They cannot be the source of that information.

We need to make sure that manufacturing information is correct and stored permanently in our manufacturing database (Light Hotspot II will also carry it on-board).

Long story short, the label has to get that information at manufacturing time from batch parameters and we shouldn't have freq in our fleets.

**Further context:**
If you ask how that label info will be kept consistent, easy. The production happens with batches. Those batches are naturally differentiated by the frequency. So while the personnel will be switching between batches, they'll make sure those parameters are correct. This way, we'll also make sure manufacturing matches to the order. A checkpoint for a proper QA. Then before closing the devices, personnel will check the FREQ of concentrator with the label. This will be another checkpoint for QC.

marvinmarnold:
@pritamghanghas @KevinWassermann94 can this be closed?

pritamghanghas:
@KevinWassermann94 @posterzh Did we ever start using nebra.json for manufacturing. I think we can merge this after that is done.

PR #394, opened 2022-06-20
Use SWARM_KEY_URI from hardware definitions

posterzh:
**Issue**

- Link: https://github.com/NebraLtd/hm-pyhelper/issues/164
- Summary:
In [detect_ecc](https://github.com/NebraLtd/hm-diag/blob/22a67fad95cd889f5ad45900657afd9a17b09dfb/hw_diag/utilities/hardware.py#L186), `KEY_STORAGE_BUS` field of hardware definitions is referenced to get the i2c bus number of ECC.
In the process of updating `gateway_mfr_rs` to the latest version v0.2.1, `SWARM_KEY_URI` field is added to the hardware definitions to specify the address of the ECC.
We need to refactor `detect_ecc()` method so that either the `SWARM_KEY_URI` or `KEY_STORAGE_BUS` field of the hardware definition can be used for specifying the ECC address.

**How**
In `detect_ecc()` method:
- Try to get the ECC address from `SWARM_KEY_URI` field.
- If above fails, try to get the ECC bus number from `KEY_STORAGE_BUS` field.
- If above fails, fall back to the ECC bus number as 1.

Added the test cases.

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/utilities/hardware.py
+++ b/hw_diag/utilities/hardware.py
@@ -1,5 +1,7 @@
+import re
 from time import sleep
 import dbus
+from urllib.parse import urlparse
 from hm_pyhelper.logger import get_logger
 from hm_pyhelper.miner_param import get_public_keys_rust
 from hm_pyhelper.hardware_definitions import variant_definitions, get_variant_attribute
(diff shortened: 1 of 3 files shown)


hw_diag/utilities/hardware.py
```diff
@@ -188,15 +189,35 @@ def detect_ecc(diagnostics):
     # It determines which value will be available for which key
     variant = diagnostics.get('VA')
 
+    i2c_bus = ''
     try:
-        key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')
+        # Example SWARM_KEY_URI: "ecc://i2c-1:96?slot=0"
+        swarm_key_uri = get_variant_attribute(variant, 'SWARM_KEY_URI')
+        pattern = 'i2c-([0-9]+)'
```
pritamghanghas, line 196: the idea behind uri was that we will be able to use urlparser. 
eg. parser = urllib.parse("ecc://i2c-1:96?slot=0")
bus_dev = parser.hostname etc for its component parts.

posterzh: yeah, it's elegant way.

hw_diag/utilities/hardware.py
```diff
@@ -188,15 +190,36 @@ def detect_ecc(diagnostics):
...
+
     except Exception as e:
+        logging.warn("Unable to lookup SWARM_KEY_URI from hardware definitions, "
+                     "Exception message: {}".format(e))
+
+    if not i2c_bus or not i2c_bus.isnumeric():
+        try:
+            # Example KEY_STORAGE_BUS: "/dev/i2c-1"
+            key_storage_bus = get_variant_attribute(variant, 'KEY_STORAGE_BUS')
+            pattern = 'i2c-([0-9]+)'
```
pritamghanghas, line 209: this is used multiple times. Make a small function. I think old method of splitting at "-" was also ok, we don't expect anything else than i2c-<number> to come here. But I don't have any problem with additional error checking.

posterzh: Inner utility method for parsing i2c_bus address looks better.

hw_diag/utilities/hardware.py
```diff
@@ -183,20 +185,44 @@ def set_diagnostics_bt_lte(diagnostics):
     return diagnostics
 
 
+def parse_i2c_bus(address):
```
pritamghanghas, line 188: can you write a comment about this function what it does like
"It takes i2c bus as input parameter and extracts the bus number and returns it. "

posterzh: Thanks. Updated!

PR #419, opened 2022-07-27
fix: TypeError: 'NoneType' object is not subscriptable #416

pritamghanghas:
* we don't want peer book data after the light hotspot switch.

**[Issue](https://github.com/NebraLtd/hm-diag/issues/416)**

- Link:https://github.com/NebraLtd/hm-diag/issues/416
- Summary: remove unncessay call to fetch_miner_data. Leaving the capability behind. Will be removed when miner client switches to gatewayrs

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


hw_diag/tasks.py
```diff
@@ -40,14 +38,6 @@ def perform_hw_diagnostics(ship=False):  # noqa: C901
     diagnostics['PK'] = public_keys['key']
     diagnostics['AN'] = public_keys['name']
 
-    # Fetch data from miner container.
-    try:
-        diagnostics = fetch_miner_data(diagnostics)
```
marvinmarnold, line 45: This was the only invocation of `fetch_miner_data`. Can you remove that method all together and related test code?

marvinmarnold (changes requested):

MuratUrsavas (approved):

hw_diag/tasks.py
pritamghanghas: It was going to be removed with light miner branch. Not sure if helium has any plans to revert on anything. But I can remove it now as well.

marvinmarnold (approved):

hw_diag/tasks.py
KevinWassermann94: There is no need for backwards compatibility at this point

pritamghanghas: done

PR #441, opened 2023-01-02
FEAT: User Auth on Diagnostics Page

robputt:
Add user auth to the local diagnostics pages...

Features
- SQLite DB with ORM Models + Alembic Migrations
- Run DB Migrations upon application start
- Login + Change Password Form Templates
- Refactor templates for better reuse via JINJA2
- Stores password as bcrypt hash
- Logs failed login attempts and locks login form if > 4 attempts in last 10 mins
- Default password entry set to Ethernet mac address of device with colons removed
- Reset password to default using device button press when on password reset page


shawaj:
@robputt the security thing is below:

Screenshot 2023-01-03 at 00 23 29

and at https://sonarcloud.io/dashboard?id=NebraLtd_hm-diag&pullRequest=441

I can mark it as safe if it is not an issue, but wasn't sure

shawaj:
also @robputt I was thinking - is there any chance we can add some brute-force protection? perhaps by rate limiting password entries?

just thinking if someone accidentally exposes this to the open internet, we should have some protection to stop people from brute forcing

robputt:
> @robputt the security thing is below:
> 
> Screenshot 2023-01-03 at 00 23 29
> 
> and at https://sonarcloud.io/dashboard?id=NebraLtd_hm-diag&pullRequest=441
> 
> I can mark it as safe if it is not an issue, but wasn't sure

@shawaj please mark this as safe in Sonarqube, we already follow this pattern in the other view and it has been accepted in Sonarqube. 👍

robputt:
> also @robputt I was thinking - is there any chance we can add some brute-force protection? perhaps by rate limiting password entries?
> 
> just thinking if someone accidentally exposes this to the open internet, we should have some protection to stop people from brute forcing

@shawaj - yep I can add this

if password_wrong > 5 tries then block login for 15 mins? 

something like this?

shawaj:
@robputt yep exactly 😊

robputt:
> @NebraLtd/tech-support @ChristopherRush once this is merged we will need a support article explaining the functionality and why we have introduced it and possibly a blog/email as well

Don't worry this is still a few days off, I am changing how it works a bit. 🤦

shawaj:
@robputt passing 🎊🎊

shawaj:
FYI @robputt 

https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=441&id=NebraLtd_hm-diag&open=AYWDOkjZ5xpLamRQHPmL

Screenshot_20230105_201818_Chrome

robputt:
> FYI @robputt
> 
> https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=441&id=NebraLtd_hm-diag&open=AYWDOkjZ5xpLamRQHPmL
> 
> Screenshot_20230105_201818_Chrome

These are indeed different tests but Sonarqube is too dumb to understand how the different @patch affects the test outcome. One overrides BALENA_DEVICE_TYPE the other test does not ;-)

shawaj:
I'll see if I can override.

It's also complaining about repeated password strings in tests

robputt:
> I'll see if I can override.
> 
> It's also complaining about repeated password strings in tests

Well that's dumb, cos it should expect password style fixtures in unit tests that are testing password functions :-\

robputt:
@shawaj ready for review I think :-)

shawaj:
@robputt you have access to sonarcloud so you should be able to override by just clicking on the link and signing in with GitHub login

.github/workflows/python-tests.yml
```diff
@@ -27,7 +27,14 @@ jobs:
...
+    - name: Move migrations to expected location
+      run: |
+        mkdir /opt/migrations
+        cp -r migrations /opt/migrations/.
+        cp -r alembic.ini /opt/migrations/.
+        sudo mkdir /var/data
+        sudo chmod 777 /var/data
     - name: Unit tests
       run: |
         pip install pytest
```
shawaj, line 40: Shall we add pytest-cov to test-requirememts and add

--cov=hw_diag --cov=bigquery --cov-fail-under=70

shawaj, line 40: Shall we add pytest-cov to test-requirememts and add

--cov=hw_diag --cov=bigquery --cov-fail-under=70

shawaj (commented):
Looks good to me.

Perhaps @MuratUrsavas or @pritamghanghas could take a look too in case I've missed anything?

shawaj:
Oh and maybe a cheeky squash/rebase @robputt 😜

.github/workflows/python-tests.yml
robputt: Generally I would agree, but I think it's out of scope of this PR. Let's raise this in a separate PR if there is a policy change regarding coverage.

pritamghanghas (commented):
Looks like for the user, only way to recover forgotten password is reflash the miner or wipe storage from balena. Does the user have capability to wipe miner on his own without raising a request to customer support.

robputt:
> Looks like for the user, only way to recover forgotten password is reflash the miner or wipe storage from balena. Does the user have capability to wipe miner on his own without raising a request to customer support.

Not currently, they would need to ask Nebra support :-(. @shawaj any thoughts how to handle this?

.github/workflows/python-tests.yml
shawaj: I mean, we have that on other repos such as config https://github.com/NebraLtd/hm-config/blob/c580b108ecacedae69b3b0c2954704646d2118a0/.github/workflows/python-tests.yml#L22

But yeah can put in a follow up ticket if that's easier

pritamghanghas:
@shawaj Here is my suggestion, not sure about security implications.
* have a forgot password button on the login screen. When the user presses it, we can take either of the following options:
   * rest to a random password and send that over email to the user.
   * check if the connection is over local lan and allow for reset of password. Someone on the local network can be reasonably trusted.

pritamghanghas:
@kashifpk will know better about identity in dashboard, but i would expect they can be identified by both uuid and mac addresses.

shawaj:
> @shawaj Here is my suggestion, not sure about security implications.
> 
> * have a forgot password button on the login screen. When the user presses it, we can take either of the following options:
>   
>   * reset to a random password and send that over email to the user.
>   * check if the connection is over local lan and allow for reset of password. Someone on the local network can be reasonably trusted.

The issue with email is that we don't have the users email and also we would have to send through a mail server which seems like overkill on the diagnostics.

Agree about the local network thing, but TBH that can probably be spoofed and could also cause issues if someone's network gets hacked. So I think more cumbersome methods that require physical access or a reflash is probably better

shawaj:
> @kashifpk will know better about identity in dashboard, but i would expect they can be identified by both uuid and mac addresses.

The main thing I'm wondering, is if a device that's already listed in the dashboard changes it's Balena uuid but still has the same Mac address and all other info, what will happen? @kashifpk any idea?

robputt:
To be merged and deployed with https://github.com/NebraLtd/hm-config/pull/225

shawaj:
> > @kashifpk will know better about identity in dashboard, but i would expect they can be identified by both uuid and mac addresses.
> 
> The main thing I'm wondering, is if a device that's already listed in the dashboard changes it's Balena uuid but still has the same Mac address and all other info, what will happen? @kashifpk any idea?

This comment still needs resolving @KevinWassermann94 but apart from that I think we are good to go @robputt so approving but should follow this up. Maybe I'll create a new issue on hm-dashboard to follow this question up

shawaj (approved):

.github/workflows/python-tests.yml
shawaj: Follow up ticket #445

shawaj:
Follow up ticket for the above... https://github.com/NebraLtd/hm-dashboard/issues/1471

kashifpk:
> The main thing I'm wondering, is if a device that's already listed in the dashboard changes it's Balena uuid but still has the same Mac address and all other info, what will happen? @kashifpk any idea?

@shawaj -AFAIK in such case our balena celery task would create a new balena record in balena table. This is being done [here](https://github.com/NebraLtd/hm-dashboard/blob/master/app/balena_helper/tasks.py#L121). This table links with the main devices table using serial number so the FK would still be valid. One problem though would be that within the devices table itself we have a balena_uuid field that never gets updated during this process so that field would contain old/invalid balena uuid.

Issue #452, opened 2023-01-15
Json endpoint always redirects to login screen even when logged in


shawaj:
https://drive.google.com/file/d/18uR5k-m_CBHcfDFFiFK0TE_F6LK58IoR/view?usp=drivesdk

See above video. Permanently redirects between json endpoint and login and never downloads the JSON

PR #458, opened 2023-01-16
Fix JSON Auth

robputt:
Fixes https://github.com/NebraLtd/hm-diag/issues/452

### DIFF
--- a/hw_diag/views/diagnostics.py
+++ b/hw_diag/views/diagnostics.py
@@ -42,7 +42,6 @@ DIAGNOSTICS = Blueprint('DIAGNOSTICS', __name__)
 
 
 @DIAGNOSTICS.route('/json')
-@cache.cached(timeout=60)
 @authenticate
 def get_diagnostics_json():
     diagnostics = read_diagnostics_file()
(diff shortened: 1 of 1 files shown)


hw_diag/views/diagnostics.py
```diff
@@ -42,16 +42,13 @@
 
 
 @DIAGNOSTICS.route('/json')
-@cache.cached(timeout=60)
```
shawaj, line 45: Can we not have caching with the auth?

hw_diag/views/diagnostics.py
```diff
@@ -42,16 +42,13 @@
...
 @authenticate
 def get_diagnostics_json():
     diagnostics = read_diagnostics_file()
     response = jsonify(diagnostics)
     response.headers.set('Content-Disposition',
                          'attachment;filename=nebra-diag.json'
                          )
-    response.headers.set('X-Robots-Tag', 'none')
```
shawaj, line 53: Not needed I guess as behind auth so can't be found anyway?

shawaj (approved):

PR #462, opened 2023-01-19
Allow changing device hostname.

robputt:
- Sets default hostname at first boot to "nebra-<modified_eth_mac>.local" e.g. "nebra-5F1CAB.local".
- Displays hostname on the diagnostics page.
- Allows user to update hostname on the device configuration page.

### DIFF
--- a/hw_diag/templates/diagnostics_page_light_miner.html
+++ b/hw_diag/templates/diagnostics_page_light_miner.html
@@ -104,5 +104,8 @@
             </div>
           </div>
         </div>
+        <p>
+          {{ device_status }}
+        </p>
       {% endif %}
 {% endblock %}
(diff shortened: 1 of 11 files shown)


hw_diag/templates/diagnostics_page_light_miner.html
```diff
@@ -63,9 +63,9 @@ <h3 class="text-center mb-4">Diagnostics Breakdown</h3>
                   {% endif %}
                 </tr>
                 <tr>
-                  <td>Hostname</td>
+                  <td>Hostname (IP Address)</td>
```
shawaj, line 66: Minor comment but maybe Local IP would be better?

I guess we could also put the Public IP on diagnostics as well but probably better on a different line?

robputt: I think we leave it like this for now, and then we can make it look pretty with the UI updates and include both local and external IPs...

shawaj: Sounds like a good plan 👌

shawaj (approved):
LGTM

Issue #468, opened 2023-01-27
Make Diagnostics production parameter aware


MuratUrsavas:
**Acceptance Criteria**
- [ ] Diagnostics has to recognize new Docker volume `/var/nebra`
- [ ] It should be able to create necessary scripts even for the devices already on field
- [ ] Diagnostics will be the primary responsible to track these environment variables

shawaj:
Related to #444

shawaj:
@MuratUrsavas we need to make sure this works for non Nebra devices. Currently it fails...

https://github.com/NebraLtd/hm-pktfwd/issues/123

Maybe for non Nebra we can just have it put a value of none in the file?

shawaj:
@MuratUrsavas I suppose the other way we could solve this is to just set all non-Nebra fleets to have a `FREQ` variable set to something like `N/A (Non Nebra Miner)`

Then we wouldn't need to make any code changes.

What do you think?

MuratUrsavas:
Actually I was expecting a mandatory FREQ parameter, but you're right. It has to be optional. Let me fix it.

shawaj:
This one just fails silently already right @MuratUrsavas ?

https://github.com/NebraLtd/hm-diag/blob/master/setenv.sh

MuratUrsavas:
@shawaj It's actually not failing. Just prompts an error if cannot reach them from both of the sources. But it continues to work if not exists.

shawaj:
Perfect

shawaj:
@MuratUrsavas @KevinWassermann94 guessing we can close this now?

PR #469, opened 2023-01-28
Added production parameter recognition feature (#468)

MuratUrsavas:
**Issue**
#468

Also improved Dockerfile for more efficient layering (combined RUN layers)

**Checklist**

- [x] Tests added
- [x] Documentation added

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,21 +2,24 @@
 
 ####################################################################################################
 ################################## Stage: builder ##################################################
-
-FROM balenalib/raspberry-pi-debian-python:bullseye-build-20221215 as builder
+FROM balenalib/raspberry-pi-debian-python:bullseye-build-20221215 AS builder
 
 ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
 
 RUN mkdir /tmp/build
-COPY ./ /tmp/build
+COPY quectel/ /tmp/build/quectel
+COPY hw_diag/ /tmp/build/hw_diag
+COPY bigquery/ /tmp/build/bigquery
+COPY requirements.txt /tmp/build/requirements.txt
+COPY setup.py /tmp/build/setup.py
+COPY MANIFEST.in /tmp/build/MANIFEST.in
 WORKDIR /tmp/build
 
 RUN \
     install_packages \
             build-essential \
-            libdbus-glib-1-dev
-
-RUN pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
+            libdbus-glib-1-dev && \
+    pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
 
 # firehose build, the tar is obtained from  quectel.
 # there is no install target in Makefile, doing manual copy
@@ -31,8 +34,7 @@ RUN make && \
 
 ####################################################################################################
 ################################### Stage: runner ##################################################
-
-FROM balenalib/raspberry-pi-debian-python:bullseye-run-20221215 as runner
+FROM balenalib/raspberry-pi-debian-python:bullseye-run-20221215 AS runner
 
 ENV PYTHON_DEPENDENCIES_DIR=/opt/python-dependencies
 
@@ -48,16 +50,6 @@ WORKDIR /opt/
 
 # Import gpg key
 COPY keys/manufacturing-key.gpg ./
-RUN gpg --import manufacturing-key.gpg
-RUN rm manufacturing-key.gpg
-
-# @TODO: Re-enable health-check once Balena supports it fully.
-# HEALTHCHECK \
-#    --interval=120s \
-#    --timeout=5s \
-#    --start-period=15s \
-#    --retries=10 \
-#  CMD wget -q -O - http://0.0.0.0:5000/initFile.txt || exit 1
 
 # Copy packages from builder
 COPY --from=builder "$PYTHON_DEPENDENCIES_DIR" "$PYTHON_DEPENDENCIES_DIR"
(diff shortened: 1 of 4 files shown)


MuratUrsavas:
Note: Please let merging to me. I'll combine it with others.

shawaj:
Did you see these?

https://sonarcloud.io/project/security_hotspots?id=NebraLtd_hm-diag&pullRequest=469&resolved=false&types=SECURITY_HOTSPOT

.hadolint.yaml
```diff
@@ -0,0 +1,2 @@
+ignored:
+    - SC1091
```
shawaj, line 2: Why do we need this?

Dockerfile
```diff
@@ -48,8 +47,6 @@ WORKDIR /opt/
 
 # Import gpg key
 COPY keys/manufacturing-key.gpg ./
-RUN gpg --import manufacturing-key.gpg
-RUN rm manufacturing-key.gpg
 
 # @TODO: Re-enable health-check once Balena supports it fully.
```
shawaj, line 51: Guess we can remove these comments. They've been here for ages and we have never used

Dockerfile
```diff
@@ -14,9 +14,8 @@ WORKDIR /tmp/build
 RUN \
     install_packages \
             build-essential \
-            libdbus-glib-1-dev
-
-RUN pip3 install --no-cache-dir --target="$PYTHON_DEPENDENCIES_DIR" .
```
pritamghanghas, line 19: actually this might make it worse. Every RUN results in a docker fs layer. apt install isn't likely to change very often while pip3 install might change more often. It is a good practice to club only related ones together.

Dockerfile
```diff
@@ -48,8 +47,6 @@ WORKDIR /opt/
 
 # Import gpg key
 COPY keys/manufacturing-key.gpg ./
-RUN gpg --import manufacturing-key.gpg
```
pritamghanghas, line 51: I think they might be needed for HPT. @kashifpk

Dockerfile
```diff
@@ -74,10 +71,21 @@ COPY --from=builder /tmp/build/alembic.ini /opt/migrations/alembic.ini
...
+    chmod 700 /usr/sbin/start_admin_session && \
+    mkdir -p /opt/nebra
 
 # Add python dependencies to PYTHONPATH
 ENV PYTHONPATH="${PYTHON_DEPENDENCIES_DIR}:${PYTHONPATH}"
 ENV PATH="${PYTHON_DEPENDENCIES_DIR}/bin:${PATH}"
 
-ENTRYPOINT ["gunicorn", "--bind", "0.0.0.0:5000", "--timeout", "300", "hw_diag.wsgi:wsgi_app"]
+# Copy environment variables startup script
+COPY setenv.sh /opt/nebra/setenv.sh
```
pritamghanghas, line 86: a single COPY will be better. May be 'COPY scripts/*.sh /opt/nebra/'. But just an opinion. You don't actually need to do it.

setenv.sh
```diff
@@ -0,0 +1,43 @@
...
+        export FREQ=${FREQ}
+        echo "FREQ is set as ${FREQ}";
+    else
+        echo "Error: Cannot set FREQ variable.";
+    fi
+else
+    echo "FREQ variable is already set as ${FREQ}.";
+    if [ ! -f ${freq_file_name} ]; then
+        echo "FREQ parameter file doesn't exist. Setting it as ${FREQ} now.";
+        echo "${FREQ}" > ${freq_file_name}
```
pritamghanghas, line 24: may be reading from file can be a function. That way we can reuse it at four different places. we can also avoid always writing it.

pritamghanghas (approved):

MuratUrsavas:
> Did you see these?
> 
> https://sonarcloud.io/project/security_hotspots?id=NebraLtd_hm-diag&pullRequest=469&resolved=false&types=SECURITY_HOTSPOT

@shawaj Yeah, but not touched as related code was written by @pritamghanghas and here for a while. Didn't want to break anything unintentionally.

.hadolint.yaml
MuratUrsavas: I've added this for an old ShellCheck failure. Actually removed that problematic code, hence can remove this file completely.

Dockerfile
MuratUrsavas: Sure, consider it done.

Dockerfile
MuratUrsavas: Well, Docker Linting was whining about it constantly. So I've gathered them. Now looks peaceful :smile:

Dockerfile
MuratUrsavas: Let's leave it as it is. Maybe someone could add another shell script in the future which should not get into that folder. It would end up there silently. I've always fan of explicit definitions.

setenv.sh
MuratUrsavas: Makes sense.

.hadolint.yaml
MuratUrsavas: Done.

Dockerfile
MuratUrsavas: Done.

setenv.sh
MuratUrsavas: Done.

.hadolint.yaml
shawaj: 👌

Dockerfile
shawaj: 🙏

shawaj:
> > Did you see these?
> > https://sonarcloud.io/project/security_hotspots?id=NebraLtd_hm-diag&pullRequest=469&resolved=false&types=SECURITY_HOTSPOT
> 
> @shawaj Yeah, but not touched as related code was written by @pritamghanghas and here for a while. Didn't want to break anything unintentionally.

@pritamghanghas what do you think? Can we solve these in an easy way? I guess `COPY .` will solve it without breaking anything?

pritamghanghas:
> > Did you see these?
> > https://sonarcloud.io/project/security_hotspots?id=NebraLtd_hm-diag&pullRequest=469&resolved=false&types=SECURITY_HOTSPOT
> 
> @shawaj Yeah, but not touched as related code was written by @pritamghanghas and here for a while. Didn't want to break anything unintentionally.

I think it was always there. I didn't specifically do the tmpfs build dir. But if COPY . /tmp/build fixes it. I don't have any issues. I don't really know why we do tmpfs build.

MuratUrsavas:
> I think it was always there. I didn't specifically do the tmpfs build dir. But if COPY . /tmp/build fixes it. I don't have any issues. I don't really know why we do tmpfs build.

Let me add it in the next build (Don't want to wait for 17 minutes right now :smile: )

shawaj:
The code smells are pretty easy to fix. Seems somewhat pointless to me but if it stops it complaining 😂

I guess it is technically the right way to do it also

shawaj:
Fixed it for you @MuratUrsavas 

You can just log in to sonarcloud and mark then as safe. Just login with your github login on there :-)

Issue #485, opened 2023-02-09
CPU temp now missing from Local Diagnostics Screen 


WayneNebra:
CPU temp now missing from Local Diagnostics Screen 

![cpu temp](https://user-images.githubusercontent.com/90446757/217823218-e196185d-ea63-41cc-bd4e-422774af1953.jpg)

shawaj:
What fleet is this on @WayneNebra ?

I think this was a temporary issue that has been resolved.

And did it show on RockPi before?

WayneNebra:
Oh sorry I should of posted more details about fleets etc. Noticed on the SB fleet this morning, funnily enough it's now gone, also on my RockPi and Outdoor that's in "Testnet" how ever just checked the **Indoor 915 and 868 RockPi** fleet and it's still missing.

I can check some more fleets and report back if you like?

shawaj:
Could you share some links to some devices that have this issue on Balena, from different fleets/variants?

WayneNebra:
Sure, here we go. 


**HELIUM-INDOOR-915-rockpi**

https://dashboard.balena-cloud.com/devices/e7eda6ff7fd0dd0d65740308e66e7f8c


**HELIUM-INDOOR-868-rockpi**

https://dashboard.balena-cloud.com/devices/063bbca77864bdd476427a900ba9282b


**HELIUM-TESTNET-ROCKPI**

https://dashboard.balena-cloud.com/devices/6d181a5a9b5f9fc313f9872e5a39c644

shawaj:
@robputt can you look at this when you can?

shawaj:
@WayneNebra @robputt is this fixed now? Can we close it?

shawaj:
https://psutil.readthedocs.io/en/latest/#psutil.sensors_temperatures

https://github.com/NebraLtd/hm-diag/blob/94b7acd3931ed5ee7535c73c7ea621ee0706050f/hw_diag/utilities/hardware.py#L339-L352

shawaj:
RockPi seems to have soc-thermal for soc and gpu-thermal for gpu

shawaj:
RockPi:
```
root@nebra-2D845E:/opt# python3
Python 3.11.1 (main, Dec  8 2022, 10:14:58) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import psutil
>>> psutil.sensors_temperatures()
[...]

PR #525, opened 2023-04-10
fix: more accurate memory usage statistics

shawaj:
As per https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory the best way to calculate available and used memory is by using the "total" and "available" metrics.

Also see https://github.com/giampaolo/psutil/blob/master/scripts/meminfo.py

Related to #485 and #522

**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/utilities/hardware.py
+++ b/hw_diag/utilities/hardware.py
@@ -349,10 +349,12 @@ def get_device_metrics():
         except Exception:
             temperature = 0
 
+    memory_used = psutil.virtual_memory().total - psutil.virtual_memory().available
+
     return {
         'cpu': psutil.cpu_percent(),
         'memory_total': psutil.virtual_memory().total,
-        'memory_used': psutil.virtual_memory().used,
+        'memory_used': memory_used,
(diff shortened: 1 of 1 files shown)


shawaj:
@robputt could we also use `from psutil._common import bytes2human` to simplify this bit as well?

https://github.com/NebraLtd/hm-diag/blob/5bc4892c690c5715a9ac3ac0b72be3269a238475/hw_diag/templates/device_info.html#L135-L147

https://github.com/giampaolo/psutil/blob/7eadee31db2f038763a3a6f978db1ea76bbc4674/psutil/_common.py#L816

shawaj:
Actually ignore my last comment @robputt as it isn't a number so wouldn't allow us to actually show the progress bar properly I guess.

Although there also appears to be an undocumented `percent` feature...

MuratUrsavas (approved):

robputt (approved):

Issue #470, opened 2023-01-31
Auto register Third party devices on Dashboard


KevinWassermann94:
Acceptance Criteria:
- Auto register devices on Dashboard when not in manufacturing mode
- Only attempt once on first boot
- Button in diagnostics with deeplink to claim device on the Dashboard

shawaj:
Endpoint is https://dashboard.nebra.com/api/v0.1/register-third-party-hotspot

Example code:

```
 import requests
 import json
 
 url = "https://dashboard.nebra.com/api/v0.1/register-third-party-hotspot"
 device = {
     "ethernet_mac_address": "00:00:00:00:00:00",
     "wifi_mac_address": "00:00:00:00:00:00",
     "frequency": "915",
     "balena_uuid": "valid uuid",
     "onboarding_address": "onboarding_address",
     "public_address": "public_address",
     "serial_number": "serial_number",
     "variant": "NEBHNT-HHRK4"
 }
 
 register = requests.post(url, data = json.dumps(device), headers = {'Content-Type': 'application/json; charset=UTF-8'})
 
 print(register.text)
```

Returns:
```
{"detail":"This device is not running on Nebra software."}

Or

{"detail":"This device is not running on Nebra software."}
```

Ref...
```
HTTP 200 OK
Allow: POST, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "name": "Register Third Party Hotspot",
    "description": "",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ]
}
```

shawaj:
After discussion with @KevinWassermann94 in the above Slack thread, thinking the best way forward is to:
- call the API endpoint automatically (only once) on first load of the diagnostics container to make sure all devices running Nebra firmware on third party devices are logged in our system ready for hooking up to dashboard (log any errors to both dashboard and in diagnostics syslog)
- add to dashboard page, or similar (maybe just on the main info page so it is front and center?) display a deep link (perhaps with a query string appended to URL https://dashboard.nebra.com/devices/#qr-modal that pre-populates the serial and Mac address in the manual entry section of the modal) and a QR code for adding the device to the dashboard
[...]

Issue #371, opened 2022-05-15
Add QR Code for onboarding the device on the dashboard


KevinWassermann94:
In addition to NebraLtd/hm-dashboard#97 we should add a QR code to the diagnostics page to onboard the device in the dashboard.
The QR code should be displayable by hovering a tooltip promoting the dashboard

shawaj:
Related to https://github.com/NebraLtd/hm-dashboard/issues/800

shawaj:
Could also add a deep link perhaps

vpetersson:
This needs to be scoped out in the next grooming.

Issue #97, opened 2021-08-09
Banner linking to the Dashboard


vpetersson:
To streamline the experience (and to boost visibility for the dashboard), we should add a CTA in the diagnostics page for claiming the device in the dashboard

Follow up: https://github.com/NebraLtd/hm-dashboard/issues/1038

**Acceptance criteria**
- UTM parameters
- Add a top full width banner
- Link to add device page
- Do not store state of whether banner is closed. It may be possible to close the banner between refreshes.

shawaj:
@NebraLtd/developers @NebraLtd/device-team @NebraLtd/tech-support I think this is something we should prioritise to push people to the dashboard.

Perhaps even having "greyed out" features such as restart, reboot, shutdown, and other dashboard specific features that then maybe launches a popup pushing people to sign up for dashboard

marvinmarnold:
@Louisreed can you work on wireframes please?

Louisreed:
![Diagnostics Page - Dashboard notification.png](https://images.zenhubusercontent.com/6240c646b62e3d4770d258f5/13a0fbbc-5f84-44d7-bfae-ecfd4df1bdbd)

KevinWassermann94:
@Louisreed That's looking good! I think we should slightly adjust the detail message though as we don't have any checks if it was claimed. Maybe something like "Use the dashboard now to gain benefits such as remote management, premium firmware and prioritized support"

marvinmarnold:
I'm not sure the horizontal divider is adding much, it seems a little cluttered.

Louisreed:
![Diagnostics Page - Dashboard notification.png](https://images.zenhubusercontent.com/6240c646b62e3d4770d258f5/9a8f51b2-ecf6-45c8-88a3-7b08a5a67e61) @KevinWassermann94 @marvinmarnold Did you want me to propose a design for the feature @shawaj mentioned about showing features greyed out?

Louisreed:
![Diagnostics Page - Dashboard notification - Modal.png](https://images.zenhubusercontent.com/6240c646b62e3d4770d258f5/3ba74f4a-862c-41b7-bc59-beb64b7997ea)

marvinmarnold:
@Louisreed when does the modal appear? I don't think both the modal and banner are necessary.

KevinWassermann94:
@Louisreed I think that adds in too much logic. I think we should only do the banner

Louisreed:
[...]

PR #562, opened 2023-04-27
feat: Auto register Third party devices on Dashboard #NEBD-2

pritamghanghas:
* automatically register third party devices on startup
* keep the registration state saved in sqlite db
* add buys loop env var that prevents diag startup for debugging

**[Issue](https://nebraltd.atlassian.net/browse/NEBD-2)**

- Link: https://nebraltd.atlassian.net/browse/NEBD-2
- Summary:

**How**

**Screenshots**

**References**

Closes: #470
Closes: #97
Closes: #371

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/utilities/dashboard_registration.py
+++ b/hw_diag/utilities/dashboard_registration.py
@@ -1,0 +1,106 @@
+import requests
+import json
+import logging
+from collections import namedtuple
+from urllib.parse import urlencode, urlunparse, urlparse
+
+from hw_diag.utilities.db import get_value, set_value
+from hw_diag.utilities.hardware import is_nebra_device
+from hw_diag.utilities.diagnostics import cached_diagnostics_data
+from hw_diag.constants import DIAG_JSON_KEYS
+
+THIRD_PARTY_MINER_REGISTRATION_URL = \
+    "https://dashboard.nebra.com/api/v0.1/register-third-party-hotspot"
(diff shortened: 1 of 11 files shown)


hw_diag/utilities/dashboard_registration.py
```diff
@@ -0,0 +1,106 @@
...
+    }
+    return device_payload
+
+
+def register_third_party_miner() -> None:
+    '''
+    tries to register a non nebra miner with dashboard
+    '''
+    try:
+        diagnostics = cached_diagnostics_data()
```
pritamghanghas, line 57: @kashifpk Can we remove this check. I think we get "400, already registered for nebra miners".  I don't really know the background of why this is a seaprate api.

MuratUrsavas (approved):
LGTM!

shawaj:
Quick question - what does it display if the miner is already claimed in dashboard but the free trial has expired and they haven't paid?

@KevinWassermann94 maybe saying about claiming the free trial doesn't make sense in this case I'm not sure

hw_diag/utilities/dashboard_registration.py
kashifpk: @pritamghanghas Haven't looked at this code before but from what I see in the code, this should just be about caching local diagnostics data and not affect dashboard registration. Better keep it in place as it's probably meant to reduce actual diagnostic runs.

PR #567, opened 2023-05-04
Add CM4 external antenna management support

ccrisan:
**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


shawaj:
@ccrisan see if you can log in to sonarcloud (click on the "1 code smell" and then log in at top right with your GitHub profile) and see if you can get access to resolve it. I think you should just be good to check

shawaj (approved):

hw_diag/utilities/hardware.py
```diff
@@ -373,6 +380,97 @@ def is_nebra_device(diagnostics: dict) -> bool:
...
+        # If we've already got device dtparams that override fleet ones, simply append ours;
+        # if there's no device override yet, we create the override by copying the fleet dtparams.
+        if device_dtparams:
+            device_dtparams.append(EXT_ANT_DTPARAM)
+        else:
+            device_dtparams = fleet_dtparams + [EXT_ANT_DTPARAM]
+    else:
+        logging.info('Disabling external antenna')
+        try:
+            device_dtparams.remove(EXT_ANT_DTPARAM)
```
pritamghanghas, line 462: this is a little strange. But there is a chance that device is left with parameters that are not in sycn with fleet even when there are no potential antenna overrides.

hw_diag/utilities/hardware.py
```diff
@@ -373,6 +380,97 @@ def is_nebra_device(diagnostics: dict) -> bool:
...
+
+
+def set_external_antenna_enabled(enabled: bool) -> None:
+    bc = balena_cloud.BalenaCloud.new_from_env()
+
+    # The device variable overrides the fleet variable, so we need to deal with both.
+    device_dtparams, device_dtparams_var_id = _get_device_dtparams_var(bc)
+    fleet_dtparams = _get_fleet_dtparams_var(bc)
+
+    if enabled:
```
pritamghanghas, line 447: the complexity of the function can be reduced by breaking out parameter preparation.

pritamghanghas (approved):
all comments are mostly suggestions. Nothing blocking.

MuratUrsavas (approved):

hw_diag/utilities/hardware.py
shawaj: @pritamghanghas do you mean that we should remove all overrides and just revert to the fleet settings once the external antenna parameter is removed?

If so, that's probably a good idea actually

ccrisan: Indeed that's a very good idea and that's the first thing that I wanted to do here. However, the `DELETE` API call returns 403 (yes, I can create and update vars but I can't delete them). Hence the workaround.

hw_diag/utilities/hardware.py
ccrisan: This is the only actual logic inside the function and it's tightly coupled with creating/updating the device config variable. Therefore it doesn't make too much sense to pull parameter preparation out, IMHO.

hw_diag/utilities/hardware.py
shawaj: How strange. 

Guess that's just another annoying quirk of Balena!

If no other way, then this seems like the only one TBH

PR #572, opened 2023-05-11
Use pyproject & friends

ccrisan:
**Issue**

- Use `pyproject.toml` to configure the entire project.
- Use `poetry` to build the package and manage deps.
- Use `ruff` for Python linting.
- Use `bandit` for code auditing.
- Separate Python linting from the rest.
- Update CI workflows accordingly.

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


.github/workflows/linter.yml
```diff
@@ -1,56 +1,33 @@
----
-###########################
-###########################
-## Linter GitHub Actions ##
-###########################
-###########################
-name: Lint Code Base
```
shawaj, line 7: as mentioned on slack, for hm-diag and hm-config where we have Dockerfiles etc we should move the python linting out into the ruff/poetry side in a separate github action, but leave the dockerfile and other linting

just noticed we have a separate hadolinter and then duplicating in here as well...maybe we just remove that hadolint and just use the shell/hadolint from this github super linter?

not sure what you think?

shawaj: also noticed in hm-config we are running bandit but not in hm-diag or hm-pyhelper. Maybe we should add?

ccrisan: Agreed. Fixed!

kashifpk (approved):

Issue #603, opened 2023-06-08
icons and other assets don't load if no working internet connection


shawaj:
As it's possible that there might be times when internet is not guaranteed (i.e. it's like a router rather than a public website) we need to store all icons, css, js, logos, images for the diagnostics locally rather than grabbing them over the internet.

So we will need to store them all locally rather than serving from cloudflare, font awesome etc

Foot example we currently are grabbing some stuff from remote sources:

https://github.com/NebraLtd/hm-diag/blob/5705da57ed8d8f3bc41edfe1c15fcdd18e55ee0f/hw_diag/templates/template_hyper.html#L14-L15

We need to remove all of these and serve locally.

Ref https://nebraltd.slack.com/archives/C024BNQ1Y6T/p1686232425142909?thread_ts=1686164871.136669&cid=C024BNQ1Y6T

Louisreed:
Working on this now - https://github.com/NebraLtd/hm-diag/tree/%23603-Static-icons-and-assets

shawaj:
Closed by #608 and #615 and #616

PR #608, opened 2023-06-09
#603 static icons and assets

Louisreed:
**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


Louisreed:
@shawaj Sonarcloud is throwing [Unexpected missing generic font family] errors (https://sonarcloud.io/project/issues) on the fontawesome fonts after making them local. Apparently it's a known issue, any recommendations or just ignore?

https://community.sonarsource.com/t/ignore-missing-generic-font-family-s4649-for-icon-fonts/57738/3

shawaj:
@Louisreed is it not because the fonts needs to be in the static/fonts folder or am I missing something here?

Louisreed:
> 

Not sure but I'll give that a try

Louisreed:
@shawaj Seems to be the same issue...

shawaj:
@Louisreed we could just add serif or sans-serif as per https://stackoverflow.com/questions/61201673/what-should-be-the-generic-font-family-for-font-awesome-fonts

Or we could just ignore the issues I guess. Does it actually work ok with it like this?

Is this what it is serving at the moment from remote?

Louisreed:
> @Louisreed we could just add serif or sans-serif as per https://stackoverflow.com/questions/61201673/what-should-be-the-generic-font-family-for-font-awesome-fonts
> 
> Or we could just ignore the issues I guess. Does it actually work ok with it like this?
> 
> Is this what it is serving at the moment from remote?

Testing the sans-serif fallback now.

Yeah if not just ignore it's not causing any issues.

Yes currently from remote.

Louisreed:
@shawaj I think that's sorted. How does it look on your end?

shawaj:
@robputt do all static assets - stylesheets, fonts, images, js etc need to be listed in MANIFEST.in?

PR #617, opened 2023-06-24
Allow users to set region.

robputt:
Allow users to set region prior to ThingsIX onboard if not already set by gateway-rs.

**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added

### DIFF
--- a/hw_diag/templates/thix_set_region.html
+++ b/hw_diag/templates/thix_set_region.html
@@ -1,0 +1,57 @@
+{% extends 'template_hyper.html' %}
+
+{% block title %}ThingsIX Region{% endblock %}
+
+{% block body %}
+    <div class="content">
+        <br/>
+        <div class="text-center">
+            <h3 class="text-center mb-4">ThingsIX Region</h3>
+        </div>
+
+        <div class="row mb-4">
+          <div class="col-12 col-lg-3 mb-4 mb-lg-0">
+          </div>
+          <div class="col-12 col-lg-6 mb-4 mb-lg-0" style="padding:20px">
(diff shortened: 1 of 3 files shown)


hw_diag/templates/thix_set_region.html
```diff
@@ -0,0 +1,57 @@
...
+                                <p class="text-danger">{{ msg }}</p>
+                            {% endif %}
+                            <form action="/thingsix/set_region" method="POST">
+                            <div class="form-group">
+                              <select class="form-select" aria-label="Default select example" id="selRegion" name="selRegion">
+                                  <option value="EU868">EU868</option>
+                                  <option value="US915">US915</option>
+                                  <option value="CN779">CN779</option>
+                                  <option value="EU433">EU433</option>
+                                  <option value="AU915">AU915</option>
```
shawaj, line 34: Do we not need to use the actual regions that packet forwarder is expecting?

robputt: Yeah, are these not the actual regions supported? :-\ I just assumed Helium, ThingsIX et al would use the same band plans as TTN for consistency...

shawaj: Sadly no. Helium has about a billion 🤣

https://github.com/NebraLtd/hm-pktfwd/blob/master/pktfwd/config/region_config_filenames.py

shawaj:
Also @robputt can the code smell be fixed at all?

shawaj (approved):

Issue #404, opened 2022-07-05
Make most fields in big query diagnostic schema as optional


pritamghanghas:
While working on [product-management#72](https://github.com/NebraLtd/product-management/issues/72), it was observed that we might miss some diagnostic reports due to schema validation of required fields.

We don't want to miss diagnostic reports if some of the current mandatory fields are malformed or missing due to hardware/software issues. Except for timestamp, there isn't any need to mark fields as required. 
In case of errors, this might lead to diag not being uploaded and thus problem being masked. Even if there is no large problem being masked, the diag upload frequency which we use to calculate uptime becomes unreliable.

Required Fields as of date that need conversion to optional:
"BN",
"ID",
"BA",
"FR",
"FW",
"VA",
"serial_number",
"PK",
"OK"
"AN",

**Acceptance Criterion**
- Make all fields except timestamp optional in `nebra-production.hotspot_diagnostics_data.diagnostics`

pritamghanghas:
@marvinmarnold This will require making changes to the table. @htsili can probably do it. I have checked. Converting previously **required** fields to **optional** is allowed by big query. I can make the device side changes after that.

htsili:
@pritamghanghas i think this is now done. Can you please check it?

pritamghanghas:
@htsili LGTM I will work on rest of it.

marvinmarnold:
We are seeing diagnostics coming through again so I think this is production already. Is any additional work needed or can this be closed @pritamghanghas ?

pritamghanghas:
@marvinmarnold No, this is in QA now. I will check a few devices from testnet tomorrow and close the ticket.

KevinWassermann94:
@pritamghanghas This must be in production already. We are getting around 470k upload per day since Jul 28th

marvinmarnold:
@pritamghanghas can you make sure this gets into production before the end of the sprint.

@KevinWassermann94 this has changes in two places - Google Cloud schema and code. Changing the cloud schema is in prod and restored uploads for most devices. The code change still needs to be deployed but will prob not have a big impact.

PR #646, opened 2023-07-09
Bump pydantic

shawaj:
**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


shawaj:
@pritamghanghas since you originally worked on this(#404 #420 ), could you take a look?

Do we need to bump the diag schema and the relevant python files to include all of the new information available in the diagnostics JSON? Or this is not necessary?

pritamghanghas (approved):

pritamghanghas:
I don't think anyone is using it. This was to upload the data to gcp bucket that triggers a function, that puts it in big query. I would say just retire  it and save billing.

pritamghanghas:
@KevinWassermann94 would know more if anyone is using it for analysis.

pritamghanghas:
may be we should just change perform_hw_diagnostics(ship=True) to perform_hw_diagnostics(ship=False)  in app.py and be done with it.

shawaj:
@pritamghanghas Kevin says it's costing next to nothing and we do sometimes use it, albeit infrequently.

Do I need to modify it to actually provide all the data that's in the json
JSON? Or how does it work?

PR #683, opened 2024-10-28
use override values even if already set

tchey290:
**Issue**

- Link:
- Summary:

**How**

**Screenshots**

**References**

**Checklist**

- [ ] Tests added
- [ ] Documentation added


shawaj:
Thanks for this contribution @tchey290 👏

I was meaning to do this but been very short on time so this is much appreciated.

